PR #518, opened 2023-01-04
chore: apply flags config with push_events

duonganhthu43:
- [x] flags config when present should include complete set of properties
- [x] proto generator shouldn't generate on_event method
- [ ]  test case for above scenario


duonganhthu43:
Replace with this [PR](https://github.com/getdozer/dozer/pull/664)

PR #631, opened 2023-01-16
chore: update proto file as matching update from dozer-types

duonganhthu43:


dozer-api/protos/internal.proto
```diff
@@ -37,17 +37,33 @@ message ApplicationDetail {
...
   repeated EndpointInfo endpoints = 6;
   string home_dir = 7;
+  Flags flags = 8;
 }
 
+message Flags {
+  bool dynamic = 1;
+  bool grpc_web = 2;
+  bool authenticate_server_reflection = 3;
+  bool push_events = 4;
```
chubei, line 47: I think we have `push_events` `3` and `authenticate_server_reflection` `4`

chubei (changes requested):

dozer-api/protos/internal.proto
duonganhthu43: right !!, Update order in `dozer-types` also

chubei (approved):

Issue #625, opened 2023-01-16
`OnEvent` is not propagating data in some scenarios. 


v3g42:
## Describe the bug
- [ ] `On Event` is not propagating events with Postgres Sample. 
- [ ] Works for eth connector but warns some events haven't been propagated.

## Environment
`cargo run --bin dozer`

## To Reproduce
1) Postgres Example
- Quick start example https://v1.getdozer.io/docs/getting_started/2.1_quickstart
- Enable `push_events` flag. 
```
flags:
  dynamic: true
  grpc_web: true
  push_events: true
  authenticate_server_reflection: false
```
- Query grpc 
```
grpc_cli call localhost:50051 dozer.generated.stocks.Stocks.on_event ''
```
- Perform some operations
```
INSERT INTO "public"."stocks"("id","ticker","date","open","high","low","close","adj_close","volume")
VALUES
(80001,E'AA',E'1999-11-22',29.551143646240234,31.473533630371094,28.65700912475586,31.473533630371094,27.06866455078125,6577800);	


INSERT INTO "public"."stocks"("id","ticker","date","open","high","low","close","adj_close","volume")
VALUES
(80002,E'AA',E'1999-11-22',29.551143646240234,31.473533630371094,28.65700912475586,31.473533630371094,27.06866455078125,6577800);


INSERT INTO "public"."stocks"("id","ticker","date","open","high","low","close","adj_close","volume")
VALUES
(80011,E'AA',E'1999-11-22',29.551143646240234,31.473533630371094,28.65700912475586,31.473533630371094,27.06866455078125,6577800);	
	
		
DELETE FROM stocks where id > 	80000;
```

## Expected behavior
Postgres `Sink` is getting updates but not propagating to gRPC `on_event`

2) ETH Example
- https://v1.getdozer.io/docs/examples/ethereum_stats
- Enable `push_events` flag. 
- Query grpc 
```
grpc_cli call localhost:50051 dozer.generated.eth_logs.EthLogs.on_event ''
```

## Expected behavior
Generally working . But i see some WARN messages some messages were not sent on some channels.

chubei:
Postgres sample is still not working

PR #637, opened 2023-01-17
fix: OnEvent is not propagating data in some scenarios with postgres

duonganhthu43:
related [issue](https://github.com/getdozer/dozer/issues/625)

### DIFF
--- a/dozer-api/src/grpc/internal/internal_pipeline_server.rs
+++ b/dozer-api/src/grpc/internal/internal_pipeline_server.rs
@@ -33,8 +33,20 @@ impl InternalPipelineServer {
         tx: Sender<PipelineResponse>,
         receiver: Receiver<PipelineResponse>,
     ) {
-        while let Ok(event_response) = receiver.recv() {
-            let _ = tx.send(event_response);
+        loop {
+            let message = receiver.recv();
+            match message {
+                Ok(message) => {
+                    let result = tx.send(message);
+                    if let Err(e) = result {
+                        warn!("Error sending message to broadcast channel: {:?}", e);
+                    }
+                }
+                Err(err) => {
+                    warn!("Message reveived error: {:?}", err);
+                    break;
+                }
+            }
         }
     }
 }
--- a/dozer-api/src/grpc/internal/internal_pipeline_server.rs
+++ b/dozer-api/src/grpc/internal/internal_pipeline_server.rs
@@ -5,14 +5,38 @@ use crate::grpc::internal_grpc::{
 };
 use crossbeam::channel::Receiver;
 use dozer_types::{crossbeam, log::info, models::app_config::Config, tracing::warn};
-use std::{net::ToSocketAddrs, pin::Pin, thread};
-use tokio::runtime::Runtime;
+use std::{net::ToSocketAddrs, pin::Pin};
+use tokio::{
+    runtime::Runtime,
+    sync::broadcast::{self, Sender},
+};
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::{codegen::futures_core::Stream, transport::Server, Response, Status};
 
 pub struct InternalPipelineServer {
     app_config: Config,
-    receiver: Receiver<PipelineResponse>,
+    receiver: broadcast::Receiver<PipelineResponse>,
+}
+impl InternalPipelineServer {
+    pub fn new(app_config: Config, receiver: Receiver<PipelineResponse>) -> Self {
+        let (tx, rx1) = broadcast::channel::<PipelineResponse>(16);
+        tokio::spawn(async move {
+            Self::setup_broad_cast_channel(tx, receiver);
+        });
+        Self {
+            app_config,
+            receiver: rx1,
+        }
+    }
+
+    fn setup_broad_cast_channel(
+        tx: Sender<PipelineResponse>,
+        receiver: Receiver<PipelineResponse>,
+    ) {
+        while let Ok(event_response) = receiver.recv() {
+            let _ = tx.send(event_response);
+        }
+    }
 }
 type ResponseStream = Pin<Box<dyn Stream<Item = Result<PipelineResponse, Status>> + Send>>;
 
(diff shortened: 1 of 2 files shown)


dozer-api/src/grpc/internal/internal_pipeline_server.rs
```diff
@@ -6,45 +6,71 @@ use crate::grpc::internal_grpc::{
...
+            receiver: rx1,
+        }
+    }
+
+    fn setup_broad_cast_channel(
+        tx: Sender<PipelineResponse>,
+        receiver: Receiver<PipelineResponse>,
+    ) {
+        while let Ok(event_response) = receiver.recv() {
+            let _ = tx.send(event_response);
```
chubei, line 37: Maybe print a log if it's an `Err`?

chubei (commented):
I'm wondering if `broadcast` is the root cause at all. We're mixing `crossbeam` channels and `tokio` channels, which is bad in tokio runtime because `crossbeam` channels are not async-aware. Two futures can run in the same thread while one is sending and another is receiving. I think we need to remove all the crossbeam channels in async code.

dozer-api/src/grpc/internal/internal_pipeline_server.rs
duonganhthu43: yah, i want to get rid of `crossbeam` also,  but it using across everywhere in `dozer-orchestrator` gonna have a huge affect on that.

`broadcast` is for solving issue multiple clients subscribe to that method. Just found out that bug: when a client is subscribe, later client subscribe won't receive any message

chubei:
> when a client is subscribe, later client subscribe won't receive any message

I'm a little confused. I thought `InternalPipelineServer` is the only one subscribing to pipeline events?

duonganhthu43:
> > when a client is subscribe, later client subscribe won't receive any message
> 
> I'm a little confused. I thought `InternalPipelineServer` is the only one subscribing to pipeline events?
 
 our `API` + `Admin` is the client of `Internal Server`

chubei:
> our `API` + `Admin` is the client of `Internal Server`

Yes, but the `receiver` in `InternalPipelineServer` is for `InternalPipelineServer` subscribing to the pipeline, not the clients subscribing to it, right?

duonganhthu43:
> > our `API` + `Admin` is the client of `Internal Server`
> 
> Yes, but the `receiver` in `InternalPipelineServer` is for `InternalPipelineServer` subscribing to the pipeline, not the clients subscribing to it, right?

Yup, it is

chubei:
> 

Oh so you just set up the extra channel to go from crossbeam to tokio right? Not for the `broadcast`.

I'm trying to understand what fixes the problem here.

chubei:
I decided not to remove `float_value` because my analysis of the bug case was wrong.

We don't have a round trip of `dozer_types::types::Field` to `prost_refect::Value` and then to `dozer_types::types::Field`. What we have are as following:

- `dozer_types::types::Record` -> `grpc::types::Recrod`. Used in common service `query` method and `CacheSink`. Implemented in `dozer_api::grpc::types_helper`.
- `grpc::types::Record` -> `prost_reflect::DynamicMessage`. Used in typed service `on_event` method. Implemented in `dozer_api::grpc::typed::helper`.
- `dozer_types::types::Record` -> `prost_reflect::DynamicMessage`. Used in typed service `query` method. Implemented in `dozer_api::grpc::typed::helper`.

The reason that we need to use different conversion paths for `query` and `on_event` is that, for `query`, API server is directly accessing the cache, getting `dozer_types::types::Record`; for `on_event`, events first get converted to `dozer_api::grpc::types` types and then sent from app server to API server.

chubei (approved):

dozer-api/src/grpc/internal/internal_pipeline_server.rs
```diff
@@ -6,45 +6,83 @@ use crate::grpc::internal_grpc::{
...
 
 #[tonic::async_trait]
 impl InternalPipelineService for InternalPipelineServer {
     type StreamPipelineRequestStream = ResponseStream;
-
     async fn stream_pipeline_request(
         &self,
         _request: tonic::Request<PipelineRequest>,
     ) -> Result<Response<ResponseStream>, Status> {
         let (tx, rx) = tokio::sync::mpsc::channel(1000);
```
v3g42, line 62: This is the problem I believe. We should have channels that we use across pipeline to websocket comparable in terms of capacity. We probably need to give internal more capacity than what pipeline is configured. Ideally this should be adjustable through configuration @chubei @duonganhthu43 what do you think ? 

When i tested locally with higher capacity, eth messages were not dropped.

v3g42:
@duonganhthu43 @chubei can you summarize what  fixes the problem here ?

dozer-api/src/grpc/internal/internal_pipeline_server.rs
chubei: Interesting. I thought we wound block on channel becoming full. I think we should find why channel capacity affects the program behaviour becuaes otherwise backpressure won't be working properly.

duonganhthu43: we could try `unbounded_channel` if capacity is the problem

chubei:
Here's my understanding:

The internal server spawns a new future for every client listening to the event stream. Let's call these futures L1, L2...

The internal server itself is a spawned future. Let's call it S1.

S1 communicates with L1 through a channel.

Problem is that we used a `crossbeam::channel` for this communication.

When the channel is full, S1 blocks until L1 receives an item from the channel. However, if it so happens that S1 and L1 are running on the same thread, the blocking of the sending of S1 implies L1 can never get the opportunity to execute receiving.

The same happens when channel is empty and L1 gets blocked on receiving while S1 is on the same thread waiting for L1 to `await`.

The solution is to switch from `crossbeam::channel` to `tokio::channel` because tokio channels are async-aware and yields execution flow when sending or receiving blocks.

duonganhthu43:
> @duonganhthu43 @chubei can you summarize what fixes the problem here ?

few issues found out: 

- Previously, the Internal server only served 1 client subscribe => so the next client subscribed won't receive any event 
- One minor bug in casting value from `Pipeline event` to Proto type `on_event`, double and float converting => Our Postgres sample have this case 
- A delay async issue when streaming event back to client => some client not received event immediately but wait for next event, then received old + new event => That's why Ethereum with so much event always works, Postgres you have to produce at least 2 events to received it

v3g42:
@chubei @duonganhthu43 Try out increasing the capacity. But i suspect it is that. As I was able to run ethereum example with no issues with the same config ? 

@chubei should we merge https://github.com/getdozer/dozer/pull/647 onto this PR and test together just in case? 

@duonganhthu43 let me know if you are succesfully able to run the sample. 

Can you also track taking the capacity of the changes through config as a separate issue ?

chubei:
#647 won't have effect on this issue. I think we can leave it there. Or merge it before this one.

duonganhthu43:
> @duonganhthu43 let me know if you are succesfully able to run the sample.

yah, able to run now in this branch

dozer-api/src/grpc/internal/internal_pipeline_server.rs
duonganhthu43: will create issue with channel capacity config

PR #638, opened 2023-01-17
fix: Generate full text index for `String` fields

chubei:


v3g42:
Can you check if this doesn't break the samples?

chubei:
> 

Both sample works

v3g42 (approved):

PR #647, opened 2023-01-17
fix: Unify typed service `query` and `on_event` record representation

chubei:
Addressing https://github.com/getdozer/dozer/pull/637#issuecomment-1385153809

### DIFF
--- a/dozer-api/src/grpc/common/service.rs
+++ b/dozer-api/src/grpc/common/service.rs
@@ -4,17 +4,16 @@ use crate::auth::Access;
 use crate::grpc::common_grpc::common_grpc_service_server::CommonGrpcService;
 use crate::grpc::internal_grpc::PipelineResponse;
 use crate::grpc::shared_impl;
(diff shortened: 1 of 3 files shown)


v3g42:
Great. To understand correctly this will use `types.proto` in both scenarios and uses the same conversion function. Is that right ? That makes sense.

chubei:
> Great. To understand correctly this will use `types.proto` in both scenarios and uses the same conversion function. Is that right ? That makes sense.

Actually there'are still 3 kinds of conversions:

- `dozer_types::types::Record` -> `grpc::types::Record`
- `dozer_types::types::Record` -> `prost_reflect::DynamicMessage`
- `grpc::types::Record` -> `prost_reflect::DynamicMessage`

These three were implemented as 3 distinct functions. This fix merely removes the third function and replaces it with the composition of the first and second.

mediuminvader:
> @mediuminvader why is it closed?
sorry, by mistake, reopen now

v3g42 (approved):

PR #652, opened 2023-01-18
chore: check name only contract is to be decoded

v3g42:


dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
```diff
@@ -1,34 +1,35 @@
...
-- authentication: !Ethereum
-    filter:
-      from_block: 0
-      addresses: []
-      topics: []
-    wss_url: "{{ETH_WSS_URL}}"
+  - authentication: !Ethereum
+      filter:
+        from_block: 0
+        to_block: 1
```
chubei, line 6: What does `to_block` 1 mean? Only fetch one block of data to save API cost?

v3g42: This test is basically fetching all historical data of ETH based on this query. by setting to_block, we are only bringing the first block. Firstly predictable test case, secondly we contain the data to a smaller data set.

chubei (approved):

dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
v3g42: Every txn has a `block_no`
https://etherscan.io/block/16431869

PR #654, opened 2023-01-18
feat: support UInt in aggregation

chloeminkyung:
[x] sum
[x] avg
[x] min
[x] max


dozer-sql/src/pipeline/aggregation/avg.rs
```diff
@@ -26,6 +26,7 @@ impl AvgAggregator {
             FieldType::Decimal => FieldType::Decimal,
             FieldType::Float => FieldType::Float,
             FieldType::Int => FieldType::Int,
+            FieldType::UInt => FieldType::UInt,
```
mediuminvader, line 29: In the case of AVG the return type can be either an integer or a float.
For Int and UInt, we can return Float by default to cover both cases

chloeminkyung: decimal -> decimal
i64 -> float
u64 -> float
float -> float

mediuminvader (approved):

PR #666, opened 2023-01-18
chore: Download prebuilt grcov from public S3 bucket and enable cache

chubei:


chubei:
Turns out cache is actually useful. Compare https://github.com/getdozer/dozer/actions/runs/3949185228/jobs/6760077359 and https://github.com/getdozer/dozer/actions/runs/3949334169/jobs/6760406448

v3g42 (approved):

PR #690, opened 2023-01-20
fix: `CONCAT()` to support multiple arguments 

snork-alt:


xudong963 (approved):
I took a look at the ticket, very neat and well tested.

chloeminkyung (approved):

PR #759, opened 2023-01-31
fix: `cargo doc -F snowflake` fails because of `include_dir`

chubei:
With this code `cargo doc -F snowflake` and `cargo test` both pass.

On `main` `cargo doc -F snowflake` fails and `cargo test` passes.


chubei:
With this code `cargo doc -F snowflake` passes and `cargo test -F snowflake` fails.

On main `cargo doc -F snowflake` fails and `cargo test -F snowflake` passes.

PR #760, opened 2023-01-31
chore: improve coverage

chloeminkyung:


dozer-core/src/storage/lmdb_storage.rs
```diff
@@ -17,7 +17,7 @@ const DEFAULT_MAX_READERS: u32 = 256;
 const DEFAULT_MAX_MAP_SZ: size_t = 1024 * 1024 * 1024;
 
 pub struct LmdbEnvironmentManager {
-    inner: Environment,
+    pub inner: Environment,
```
chubei, line 20: This field is not `pub` on purpose, because once it's `pub`, users can use it to create transactions and use them without proper locking, which can cause data races. Users should always rely on `create_txn` to create new transactions.

chloeminkyung: got it will remove the test then

chubei: Is this design blocking you from implementing a test? Which one?

chubei: Please do this `let mut tx = storage.create_txn().unwrap();`

chloeminkyung: it's ok - just a simple unit test in the `aggregation_tests` but fixed it using new Environment attribute ;)

dozer-sql/src/pipeline/aggregation/tests/aggregation_tests.rs
```diff
@@ -68,11 +83,12 @@ fn test_simple_aggregation() {
...
     processor
-        .init(storage.as_environment())
+        .init(storage.borrow_mut())
         .unwrap_or_else(|e| panic!("{}", e.to_string()));
 
-    let binding = Arc::new(RwLock::new(storage.create_txn().unwrap()));
-    let mut tx = SharedTransaction::new(&binding);
+    let binding =
+        LmdbExclusiveTransaction::new(Environment::new().open(Path::new("./")).unwrap()).unwrap();
+    let mut tx = SharedTransaction::new(binding);
```
chubei, line 91:
```suggestion
    let mut tx = storage.create_txn().unwrap();
```

chubei (changes requested):

chubei (commented):
Only reviewed lmdb storage related part.

chubei (approved):

mediuminvader (approved):

PR #768, opened 2023-02-01
fix: Remove debugging code that sets channel capacity to 1

chubei:
Before

`cargo test -p dozer-core test_checpoint_consistency_ns  4.66s user 2.53s system 358% cpu 2.008 total`

After

`cargo test -p dozer-core test_checpoint_consistency_ns  2.30s user 0.16s system 242% cpu 1.014 total`

We really need to keep commit sizes small so this kind of bug doesn't slip from the eye of the reviewer.


chubei:
Included in #729

PR #775, opened 2023-02-01
feat: Implement basic continue for ingestion in snowflake

karolisg:


dozer-orchestrator/src/pipeline/sinks.rs
```diff
@@ -389,6 +389,7 @@ impl Sink for CacheSink {
                     })?;
             }
             Operation::Insert { mut new } => {
+                info!("NEW RECORD: {:?}", new);
```
mediuminvader, line 392: do we need this?

PR #800, opened 2023-02-03
fix: Fix record delete when duplicated words are in string field

### LINKED ISSUE #799: Full text indexing crashes when duplicated words are in string
## Describe the bug
The secondary indexes cache fails with error `Failed to delete a record - NotFound` when there is field with duplicated words.

## Environment
- **Dozer version**: 0.1.6

## To Reproduce
- Have string field in schema.  
- Use duplicated words in string. For example `regular test regular`
- Try to delete record

## Expected behavior
Indexer shouldn't try to delete duplicated words twice

karolisg:
Fixes #799

### DIFF
--- a/dozer-cache/src/cache/lmdb/indexer.rs
+++ b/dozer-cache/src/cache/lmdb/indexer.rs
@@ -5,6 +5,7 @@ use dozer_types::{
 };
 use lmdb::{RwTransaction, Transaction};
 use std::sync::Arc;
+use itertools::Itertools;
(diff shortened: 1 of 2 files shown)


chubei (approved):

chubei:
I think on insertion we need that `unique` too. I'm just making a mental note here and will fix later.

chubei (approved):
Oh just realized that function is called on both insertion and deletion

PR #801, opened 2023-02-03
refactor: Don't try to get schema of unconnected ports

chubei:
This PR merges `Dag` and `DagSchema` to a single struct `DagSchema`, and keeps the schema information on the edges, for convenient future use.

A side effect is that, now if an output port is not used (not connected to any input port), we won't try to get its schema.

I'm not sure if it's the behaviour we want? @snork-alt

### DIFF
--- a/dozer-core/src/dag/dag_impl.rs
+++ b/dozer-core/src/dag/dag_impl.rs
@@ -1,4 +1,4 @@
-use daggy::petgraph::visit::{Bfs, EdgeRef, IntoEdges, Topo};
+use daggy::petgraph::visit::{Bfs, EdgeRef, IntoEdges, IntoNodeReferences};
 use daggy::Walker;
 
 use crate::dag::errors::ExecutionError;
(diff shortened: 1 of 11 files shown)


snork-alt:
> This PR merges `Dag` and `DagSchema` to a single struct `DagSchema`, and keeps the schema information on the edges, for convenient future use.
> 
> A side effect is that, now if an output port is not used (not connected to any input port), we won't try to get its schema.
> 
> I'm not sure if it's the behaviour we want? @snork-alt

I think it is okay. Anyway if the port is not connected we will not be propagating the schema right ?

chubei:
The schema cannot be propagated yes. The difference is that we used to call `source/processor.get_output_schema` with an unconnected port. Now we don't even call that function. I don't know if the call has side effects.

snork-alt:
> The schema cannot be propagated yes. The difference is that we used to call `source/processor.get_output_schema` with an unconnected port. Now we don't even call that function. I don't know if the call has side effects.

It shouldn't have any side effects. The only reason was to propagate the schema to the next node

snork-alt (approved):

PR #818, opened 2023-02-07
chore: fix dag issues

v3g42:


dozer-core/src/dag/executor.rs
```diff
@@ -166,43 +167,75 @@ impl<'a, T: Clone + Debug + 'static> DagExecutor<'a, T> {
...
+            return Err(IncompatibleSchemas(
+                "Input Schemas length mismatch".to_string(),
+            ));
         }
-        for (port, (schema, _ctx)) in &current.output_schemas {
-            let other_schema = existing
-                .output_schemas
-                .get(port)
-                .ok_or(IncompatibleSchemas())?;
+        for (port, (schema, _ctx)) in &current.input_schemas {
```
chloeminkyung, line 217: should be `output_schemas`?

chubei: I'm not sure. @v3g42 said it was a fix.

v3g42: No. I am not sure about this code actually. I saw two code blocks that repeated. I assumed one was for input_schemas. Can you check the logic properly again ?

chubei: This commit's version is correct. Resovling this conversation.

chloeminkyung (approved):

PR #821, opened 2023-02-07
refactor: Make `LmdbRoCache` and `LmdbRwCache` `Send` and `Sync`

chubei:
There're a lot of changed lines but most of them are routine changes to make the code compile.

The core changes are that `trait Cache` is spliited to `trait RoCache` and `trait RwCache`. The `insert`, `update` and `delete` of `RwCache` no longer commit immediately. Users have to call `commit` to persistent the change.

The corresponding implementations are `LmdbRoCache` and `LmdbRwCache`. They both use the environment and the exclusive transaction from `dozer-storage` now, which makes them `Send` and `Sync`.


v3g42:
@chubei THis will be great if it works seamlessly. Can you test with today's scenario and the samples just in case  ?

chubei:
> @chubei THis will be great if it works seamlessly. Can you test with today's scenario and the samples just in case ?

Sure

snork-alt (approved):

chubei:
All three samples are working.

PR #823, opened 2023-02-07
chore: update with reformatted config

chloeminkyung:


examples/1_hypercharge_postgres/dozer-config.yaml
```diff
@@ -41,14 +61,13 @@ sources:
 endpoints:
   - name: stocks
     path: /stocks
-    table_name: stocks
```
v3g42, line 44: why dont you let one API connect directly to source but select the columns at the source level and another transformed using SQL. 

And put comments near the endpoints explaining how they work ?

examples/2_eth_stats_sample/dozer-config.yaml
```diff
@@ -23,13 +23,26 @@ connections:
     name: eth_conn
 
 sql: |
-  select punkIndex, minValue, toAddress
-  INTO punk_offered from punks_PunkOffered;
+  SELECT
```
v3g42, line 26: I think we should add comments for each of these too. So users can understand when they see it.

examples/1_hypercharge_postgres/dozer-config.yaml
chloeminkyung: sure - will do

examples/2_eth_stats_sample/dozer-config.yaml
chloeminkyung: ok

examples/1_hypercharge_postgres/dozer-config.yaml
```diff
@@ -38,17 +38,31 @@ sources:
...
     index:
       primary_key:
         - id
-  - id: null
-    name: stocks_meta
+  - name: stocks_meta
     path: /stocks-meta
     table_name: stocks_meta
+    # direct selecting columns at the source level for `stocks_meta`
+    sql: SELECT * from stocks_meta;
```
v3g42, line 65: I m guessing this will fail ? There is no SQL property on endpoins anymore.

chloeminkyung: addressed

v3g42 (approved):

PR #830, opened 2023-02-08
refactor: Remove unused connector trait method

karolisg:


chloeminkyung (approved):

v3g42:
@karolisg On second thought, shoudl we remove get_schemas and leave get_tables instead ? Basically whicher has information sufficient for both cases.

karolisg:
> @karolisg On second thought, shoudl we remove get_schemas and leave get_tables instead ? Basically whicher has information sufficient for both cases.

@v3g42 `get_tables`is supposed to return all tables and columns, while `get_schemas` is only used to map selected tables 
with only selected columns to our schema. It might not be very efficient to fetch all tables and map them to schemas on every app run/migration.

v3g42:
I know the current way it is implemented doesn't make sense.

What I also want to know typically is 
- Names of the tables
- Field Definitions
- Indexes defined on these tables. 

Querying all vs select tables should be a property on the method and it is a user's choice. For example, I want to first view all the available tables and definitions before I know what tables to use. 

```
 fn get_tables(&self) -> Result<Vec<TableInfo>, ConnectorError> {
        self.schema_helper.get_tables(None)
    }

    fn get_schemas(
        &self,
        table_names: Option<Vec<TableInfo>>,
    ) -> Result<Vec<SchemaWithChangesType>, ConnectorError> {
        self.schema_helper
            .get_schemas(table_names)
            .map_err(ConnectorError::PostgresConnectorError)
    }
```

PR #831, opened 2023-02-08
chore: Improve Join processor errors

mediuminvader:


dozer-sql/src/pipeline/product/processor.rs
```diff
@@ -48,14 +48,16 @@ impl FromProcessor {
...
+        self.operator
+            .execute(
+                JoinAction::Delete,
+                from_port,
+                record,
+                database,
+                transaction,
+                reader,
+            )
+            .map_err(|err| ExecutionError::ProductProcessorError(Box::new(err)))
```
v3g42, line 60: Do these errors have a context? Like only happening in insert etc. If so, I would do a transparent on `ExecutionError::ProductProcessorError` and specify specific situations in ProductProcessorError:Insert` etc

Not sure if this is the case.

dozer-sql/src/pipeline/product/join.rs
```diff
@@ -840,30 +852,43 @@ impl JoinOperator {
...
+            let entry = cursor
+                .read()
+                .map_err(|err| ProductError::IndexGetError(join_key.to_vec(), err))?;
+
+            if entry.is_none() {
+                break;
+            }
 
-            if entry.0 != join_key {
+            let (key, value) = entry.unwrap();
```
v3g42, line 881: are you sure this unwrap wont cause issues ?

mediuminvader: I'm doing the check at line 877 do you think 
```suggestion
            if let Some(key, value) = entry {
```
would be more readable?

snork-alt (approved):

PR #832, opened 2023-02-08
chore: Remove unused struct `OperationEvent`

chubei:


dozer-ingestion/src/connectors/snowflake/tests.rs
```diff
@@ -48,9 +47,7 @@ fn connector_disabled_test_e2e_connect_snowflake_and_read_from_stream() {
         let op = iterator.write().next();
```
chubei, line 47: @karolisg This test looks weird. I didn't change the behaviour but the old behaviour doesn't seem right.

snork-alt (approved):

dozer-ingestion/src/connectors/snowflake/tests.rs
chubei: From @karolisg the test will hang if we can't get data from snowfake.

PR #852, opened 2023-02-10
chore: Remove unused connector function

chubei:


karolisg (commented):
`test_connection` is supposed to be used in first step of connection creation on admin UI. It isn't a good idea to remove it from trait, because later we will need to add it again.

chubei:
> `test_connection` is supposed to be used in first step of connection creation on admin UI. It isn't a good idea to remove it from trait, because later we will need to add it again.

I thought that functionality is implemented in validate. So these two functions have different purposes?

karolisg:
> > `test_connection` is supposed to be used in first step of connection creation on admin UI. It isn't a good idea to remove it from trait, because later we will need to add it again.
> 
> I thought that functionality is implemented in validate. So these two functions have different purposes?

Yeah, it is different.

Basically, UI was supposed to have two steps - add connection and add source.
To validate `add connection` we were using `test_connection` function because we were only checking for connection properties.

Anyway, I double check current implementation. Probably it is safe to remove `test_connection` and later we can adjust `validate` function if needed.

karolisg (approved):

PR #907, opened 2023-02-15
refactor: Don't go through `serde_json::Value` during serialization

chubei:
We allowed more query expressions, such as arbitary field names, object and array as filter value,
empty and single $and filter and nested $and filter.


chubei:
Included in #858

Issue #887, opened 2023-02-14
feat: Geospatial types support (initial version - `POINTS` only)


mediuminvader:
### Discussed in https://github.com/getdozer/dozer/discussions/877

<div type='discussions-op-text'>

<sup>Originally posted by **snork-alt** February 13, 2023</sup>
We need to support geospatial types and functions in Dozer. The work will consist of the following:

- Extend `dozer-types` to support geospatial types
- Allow creation and casting of geospatial types
- Include support in connectors for geospatial-types conversions
- Include support for geospatial types in the cache writer
- Implement various geospatial functions
- Update all relevant documentation

For the implementation, we will use the Rust Geo library (https://docs.rs/geo/latest/geo/)
The initial version will have limited support for geo data types. We will extend it later after we will introduce the ARRAY type

### Types support

We shall extend the `dozer-types` with the following types mapped 1 to 1 with the Rust Geo types:

- `Coord`
- `Point`
-  `Polygon`

## Creation

The creation of the above types should be supported through SQL functions. In particular:

- COORD(FLOAT, FLOAT)
- POINT(COORD)
- POLYGON(COORDS, ...)

## Functions

DISTANCE(POINT, POINT, [GEODESIC | HAVERSINE | VINCENTY]) - defaults to GEODESIC
CONTAINS(POLYGON, POINT | POLYGON)
</div>

PR #912, opened 2023-02-15
feat: Add geo point and coord type support

karolisg:
Part of #887 

- [x] Add type
- [x] postgres mapping
- [ ] snowflake mapping
- [x] tests
- [ ] GRPC support
- [x] Ord, PartialEq, PartialOrd implementation for DozerPoint and DozerCoord
- [x] SQL `POINT`
- [x] SQL `DISTANCE`

### DIFF
--- a/dozer-api/protos/types.proto
+++ b/dozer-api/protos/types.proto
@@ -58,6 +58,8 @@ enum Type {
   Timestamp = 8; // ISO 8601 combined date and time with time zone.
   Date = 9; // ISO 8601 calendar date without timezone.
   Bson = 10; // BSON data.
+  Coord = 11; // Geo Coord type.
+  Point = 12; // Geo Point type.
 }
 message SchemaEvent {
   string endpoint = 1;
@@ -76,6 +78,10 @@ message FieldDefinition {
   bool nullable = 3;
 }
 
+message CoordType {
+  map<string, float> values = 1;
+}
+
 // A field value.
 message Value {
   // A field value.
@@ -88,6 +94,7 @@ message Value {
     bytes bytes_value = 7; // Binary data.
     ArrayValue array_value = 8; // Value array.
     double double_value = 9; // 64 bit floating point number.
+    CoordType map_value = 10;
   };
 }
 
--- a/dozer-api/protos/types.proto
+++ b/dozer-api/protos/types.proto
@@ -58,8 +58,7 @@ enum Type {
   Timestamp = 8; // ISO 8601 combined date and time with time zone.
   Date = 9; // ISO 8601 calendar date without timezone.
   Bson = 10; // BSON data.
-  Coord = 11; // Geo Coord type.
-  Point = 12; // Geo Point type.
+  Point = 11; // Geo Point type.
 }
 message SchemaEvent {
   string endpoint = 1;
@@ -78,8 +77,9 @@ message FieldDefinition {
   bool nullable = 3;
 }
 
-message CoordType {
-  map<string, float> values = 1;
+message PointType {
+  double x = 1;
+  double y = 2;
 }
 
 // A field value.
@@ -94,7 +94,7 @@ message Value {
     bytes bytes_value = 7; // Binary data.
     ArrayValue array_value = 8; // Value array.
     double double_value = 9; // 64 bit floating point number.
-    CoordType map_value = 10;
+    PointType map_value = 10;
   };
 }
 
--- a/dozer-api/protos/types.proto
+++ b/dozer-api/protos/types.proto
@@ -94,7 +94,7 @@ message Value {
     bytes bytes_value = 7; // Binary data.
     ArrayValue array_value = 8; // Value array.
     double double_value = 9; // 64 bit floating point number.
-    PointType map_value = 10;
+    PointType point_value = 10;
   };
 }
 
--- a/dozer-sql/src/pipeline/expression/geo/tests/distance.rs
+++ b/dozer-sql/src/pipeline/expression/geo/tests/distance.rs
@@ -1,0 +1,176 @@
+use crate::pipeline::errors::PipelineError::{
+    InvalidFunctionArgumentType, NotEnoughArguments, TooManyArguments,
+};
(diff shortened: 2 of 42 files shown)


dozer-types/src/types/mod.rs
```diff
@@ -238,3 +242,103 @@ pub enum Operation {
     Insert { new: Record },
     Update { old: Record, new: Record },
 }
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+pub struct DozerCoord(pub Coord);
+
+impl Eq for DozerCoord {}
```
chubei, line 249: Instead of implementing ordering traits, we can do this:

```rust
#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
pub struct OrderedCoordFloat(pub OrderedFloat<f64>);

impl geo::CoordFloat for OrderedCoordFloat {}

#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
pub struct DozerCoord(pub Coord<OrderedCoordFloat>);
```

dozer-types/src/types/mod.rs
```diff
@@ -238,3 +242,103 @@ pub enum Operation {
...
+    }
+}
+
+impl DozerCoord {
+    pub fn as_bytes(&self) -> Vec<u8> {
+        let mut x_bytes_vec = self.0.x.encode_to_vec();
+        let mut y_bytes_vec = self.0.y.encode_to_vec();
+        x_bytes_vec.append(&mut y_bytes_vec);
+        x_bytes_vec
+    }
```
chubei, line 287: `as_bytes` usually returns a `&[u8]`. Let's call it `to_bytes`.

We can change it like this

```rust
pub fn to_bytes(&self) -> [u8; 16] {
  let mut result = [0; 16];
  result[0..8].copy_from_slice(self.0.x.to_be_bytes());
  result[8..16].copy_from_slice(self.0.y.to_be_bytes());
  result
}
```

to avoid allocations.

chubei (changes requested):

dozer-api/protos/types.proto
```diff
@@ -58,6 +58,8 @@ enum Type {
   Timestamp = 8; // ISO 8601 combined date and time with time zone.
   Date = 9; // ISO 8601 calendar date without timezone.
   Bson = 10; // BSON data.
+  Coord = 11; // Geo Coord type.
```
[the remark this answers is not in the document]
karolisg: Removed. Thanks for the suggestion.

dozer-types/src/types/field.rs
```diff
@@ -105,10 +117,27 @@ impl Field {
...
     }
 
+    pub fn decode_x_y(bytes: &[u8]) -> Result<(f64, f64), DeserializationError> {
+        let x = f64::from_be_bytes(
+            bytes[0..8]
+                .try_into()
+                .map_err(|_| DeserializationError::BadDataLength)?,
+        );
+        let y = f64::from_be_bytes(
+            bytes[9..17]
```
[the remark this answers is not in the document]
karolisg: Right. Thanks.

dozer-types/src/types/mod.rs
karolisg: I have removed `DozerCoord`, which your suggestion was referring to.

Anyway, I have tried to use a similar approach for `DozerPoint`, but to use OderedFloat, we need to implement a bunch of methods from supertrait of `CoordFloat`.

chubei: I'm afraid that's necessary. Implementing `Eq` for `f64` is unsound.

karolisg:
@mediuminvader GRPC support and snowflake type mapping with our point type should be implemented in separate PR

dozer-types/src/types/field.rs
```diff
@@ -44,6 +48,7 @@ impl ToPyObject for Field {
             }
             Field::Bson(val) => val.to_object(py),
             Field::Null => unreachable!(),
+            Field::Point(_val) => todo!(),
```
chubei, line 51: @xudong963 can you check this and suggest what to do?

xudong963: Leave it as todo, after the ticket merged, I'll open a ticket to complete it.

dozer-types/src/types/mod.rs
```diff
@@ -331,3 +340,91 @@ pub enum Operation {
...
+    pub fn to_bytes(&self) -> [u8; 16] {
+        let mut result = [0_u8; 16];
+        result[0..8].copy_from_slice(&self.0.x().to_be_bytes());
+        result[8..16].copy_from_slice(&self.0.y().to_be_bytes());
+        result
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> Result<(f64, f64), DeserializationError> {
+        let x = f64::from_be_bytes(
+            bytes[0..7]
```
chubei, line 418: Should be `[0..8]`? And is it better to just return `Self`? Like `pub fn from_bytes(bytes: [u8; 16]) -> Self` and let the length error be handled elsewhere?

karolisg: Thanks. Fixed.

chubei (approved):
Looks good to me. Didn't check `dozer-sql` related part. @mediuminvader Do you want to have a look?

mediuminvader (approved):
We need to open new issues for SNOWFLAKE support

dozer-sql/src/pipeline/expression/geo/tests/distance.rs
```diff
@@ -0,0 +1,173 @@
...
+            _expected_types,
+            1
+        ))
+    ));
+}
+
+#[test]
+fn test_distance() {
+    let tests = vec![
+        ("", 1113.0264976969002),
```
chubei, line 102: This test doesn't pass on my Apple Silicon machine. I think we need to test with epsilon tolerance. What do you think @karolisg ?

PR #913, opened 2023-02-15
Reduce allocations & clones. Minor improved ergonomics

auterium:
- Refactored how the `fields` vector is built to avoid reallocations
- Extracted some duplicated code into functions to DRY
- Replaced `map_or_else()` for `ok_or()` to reduce verbosity
- Reduced verbosity of errors

### DIFF
--- a/dozer-ingestion/src/connectors/object_store/table_reader.rs
+++ b/dozer-ingestion/src/connectors/object_store/table_reader.rs
@@ -55,8 +55,8 @@ impl<T: Clone + Send + Sync> TableReader<T> {
                 .map_err(ObjectStoreConnectorError::InternalDataFusionError)?,
         );
 
-        let columns: Vec<ColumnInfo> = match table.columns.clone() {
-            Some(columns_list) if !columns_list.is_empty() => columns_list.to_vec(),
+        let columns: Vec<ColumnInfo> = match &table.columns {
+            Some(columns_list) if !columns_list.is_empty() => columns_list.clone(),
             _ => fields
                 .iter()
                 .map(|f| ColumnInfo {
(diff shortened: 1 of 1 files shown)


dozer-ingestion/src/connectors/object_store/table_reader.rs
```diff
@@ -77,38 +77,33 @@ impl<T: Clone + Send + Sync> TableReader<T> {
...
-                        .unwrap();
-
-                    idx += 1;
-                }
-            } else {
-                break;
+
+        while let Some(Ok(batch)) = data.next().await {
+            for row in 0..batch.num_rows() {
+                let fields = batch
```
auterium, line 83: Main impacting change being here, as this `iter() + collect()` approach will pre-allocate all the necessary memory for the resulting vector instead of starting with an empty vector that causes multiple re-allocations as elements are pushed to it.

Additionally, the `.columns()` method allows for direct access to the slice, so it's simpler code without extra overhead (maybe even slightly more performant)

dozer-ingestion/src/connectors/object_store/table_reader.rs
```diff
@@ -77,38 +77,33 @@ impl<T: Clone + Send + Sync> TableReader<T> {
...
-                            }),
-                        ))
-                        .unwrap();
-
-                    idx += 1;
-                }
-            } else {
-                break;
+
+        while let Some(Ok(batch)) = data.next().await {
```
auterium, line 81: As far as I know there's no difference between this format and the `loop + break` approach and it reduces complexity/verbosity so I thought it would be OK to change. If there's preference for the other format, I have no problem on reverting to it

chubei (approved):
LGTM. The ci failure is probably irrelevant. That test do this sometimes. I reran it.

auterium:
> LGTM. The ci failure is probably irrelevant. That test do this sometimes. I reran it.

Thanks! I've got other improvements on this part of the code that I'd like to propose as well, but wanted to start small 😃 Once this PR is merged I'll work on the next one

mediuminvader (approved):

mediuminvader:
Thanks @auterium

PR #916, opened 2023-02-16
refactor: Cleanup `running` flag and excessive `Ingestor` cloning

chubei:
This PR does two things (it's better if it's two PRs, but I somehow ended up with all the changes in one commit):

## Running flag

This PR removes `running.store` in source sender thread and in the unused `stop` method, and  `running.load` when app threads (connector, source listener, processor) quit with an error.

Now there's only one place changing the flag from false to true in the `ctrlc` handler, and only one place reading that flag in source listener.

## Ingestor

Now that `Ingestor` and `IngestionIterator` are effectively two sides of a single producer single consumer channel, there's no need to wrap them in `Arc<RwLock<_>>` and copy them around.

## Style changes

- Removed `Connector::initialize` because it's always called just before `start`. Merging the two methods to one saves the connectors from storing the `ingestor` and `tables` in `initialize` and then using them in `start`.
- Removed `internal_err` macro because macros are less readable and prevent VSCode from showing compilation errors in correct file.

This is a draft because the ingestor refactor breaks an ignored unit test `single_source_sink`. You can go ahead and review while I fix that test.


chubei:
Actually `EventsConnector` won't work as is since we've switched to static schema. Maybe we can remove the connector and test all togegher @snork-alt ?

snork-alt (approved):

chubei:
@karolisg Do you want to have a look at this?

karolisg:
LGTM

PR #918, opened 2023-02-16
ObjectStore - Reduce code duplication & more clone prevention

auterium:
- Proposed trait `DozerObjectStore` that defines a method that provides the necessary pieces to connect to a store which allows for separating the logic of "what to do" from "how to do it" on the connection to the object stores. If accepted, this might even be used in the `SchemaMapper` as well
- Slight change in the `Reader` definition as it was expecting owned values to be passed when in practice they're not needed
- Slight change in `map_schema_to_dozer()` to receive an `Iterator` which allows for on-the-fly filtering instead of cloning subsets of tables


chubei:
I don't understand object store enough to review this. @karolisg Can you have a look?

karolisg (approved):

mediuminvader (approved):

PR #921, opened 2023-02-16
Add missing types.proto reference

auterium:
When trying to run the code base it compiles but it fails to run as the `types.proto` file is not copied to the `.dozer/api/generated` folder


chubei:
I need to look into this further. I think types.proto was put into the generated folder by some other code. Let me confirm if it's some other bug before merging this fix.

auterium:
@chubei no problem. The reason I had to put it was because it's used by `common.proto`

chubei (approved):
This is indeed needed when the config file has no `endpoints`. Thanks!

chubei:
Hi @auterium were you using a config file without endpoints when you saw this error? Because I wanted to add a test case for the no endpoint case but a `debug_assert!` prevents me from doing it. I'm afraid I might have made a mistake identifying the root cause.

PR #927, opened 2023-02-17
feat: support python udf

xudong963:
- [x] Make dozer core run python udf
- [x] Add tests
- [x] install virtualenv in ci to run python udf test.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -53,10 +53,6 @@ jobs:
           restore-keys: |
             coverage-${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
             coverage-${{ runner.os }}-cargo-
-      - name: Create python virtual env
-        shell: bash
-        run: |
-          source ./dozer-tests/python_udf/virtualenv.sh
       - name: MongoDB in GitHub Actions
         uses: supercharge/mongodb-github-action@1.8.0
 
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -62,14 +62,15 @@ jobs:
 
       - uses: ./.github/workflows/setup-snowflake-and-kafka
 
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --all-features --no-fail-fast
+      - name: Cargo test
         env:
           CARGO_INCREMENTAL: "0"
           RUSTFLAGS: "-Cinstrument-coverage"
           LLVM_PROFILE_FILE: "cargo-test-%p-%m.profraw"
+        shell: bash
+        run: |
+          source ./dozer-tests/python_udf/virtualenv.sh
+          cargo test --all-features --no-fail-fast
 
       - uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -53,7 +53,10 @@ jobs:
           restore-keys: |
             coverage-${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
             coverage-${{ runner.os }}-cargo-
-
+      - name: Create python virtual env
+        shell: bash
+        run: |
+          source ./dozer-tests/python_udf/virtualenv.sh
       - name: MongoDB in GitHub Actions
         uses: supercharge/mongodb-github-action@1.8.0
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1990,6 +1990,7 @@ dependencies = [
  "parking_lot",
  "prettytable-rs",
  "prost",
+ "pyo3",
  "rust_decimal",
  "serde",
  "serde_json",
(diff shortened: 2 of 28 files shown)


dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
```diff
@@ -12,7 +12,7 @@ connections:
     name: eth_logs
 sql: |
   -- Eth stats table
-  select block_number, sum(id) into eth_stats
+  select block_number, sum(id), py_udf('add', id) into eth_stats
```
snork-alt, line 15: Can we have a syntax `py:add()`

dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
```diff
@@ -43,3 +43,10 @@ endpoints:
       primary_key:
         - block_number
         - id
+
```
snork-alt, line 46: Can we avoid having to define udfs in the config, but rather Discovery them from the .py file itself ?

dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
xudong963: No problem

dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
xudong963: If not defined udfs in config, should we require users to put .py file in a specific location?  cc @mediuminvader

snork-alt: Would be nice to have a python decorator

dozer-tests/src/e2e_tests/cases/eth/dozer-config.yaml
xudong963: `:` or `::` will be handled incorrectly by sqlparser and return unintended results. If we still want to use `:`, I'll check sqlparser deeply

xudong963:
Update the status:

Now the ticket works by the following steps:

1. Create virtualenv, such as `virtualenv test`, and activate it: `./test/bin/activate`
2. Create module `py_udf` (currently the module is fixed) in `test`
3. Add some functions in `py_udf`, such as: 
    ```python
   def add(a):
     return a + 1


    def sum(a, b):
      return a + b
    ```
4. Use python udf in dozer, such as `SELECT py_add(a, 'FLOAT'), py_sum(a, b, 'FLOAT') from t1;` (The last arg in function is return type).

Table t1, with a, b field, and values 1, 2, will get the results:
image


Question: how should we test the ticket? I think current sql test framework can't test python udf because sqlite doesn't the sql syntax.

My thought is we can change the sql test framework and make it accept self-defined`source_result`.

snork-alt:
Feel free to modify the test framework to test python execution. Instead of having to specify the return type during the invocation of the function, can we specify as an annotation in the definition of the function itself? Something like:

```
@dozer_type(Int)
def add(a):
 return a + 1
```

xudong963:
> Instead of having to specify the return type during the invocation of the function, can we specify as an annotation in the definition of the function itself? Something like:

Do you mean let users write `@dozer_type(Int)` in their python file function definition ahead? If so, I'd more like the current way.

Imagine in users' python files

```python
@dozer_type(Int)
def ...

@dozer_type(..)
def ...

...

```
This can cause a certain level of intrusion on the user side, and it's not good for us to get this information.

For the current way, the user only needs to know our constraint (the last parameter is the return type), it is easy to write the return type to the sql, and we can easily get this information only during the parse process, the intrusion to our code is almost zero.

.github/workflows/coverage.yaml
```diff
@@ -53,7 +53,10 @@ jobs:
           restore-keys: |
             coverage-${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
             coverage-${{ runner.os }}-cargo-
-
+      - name: Create python virtual env
+        shell: bash
+        run: |
+          source ./dozer-tests/python_udf/virtualenv.sh
```
xudong963, line 59: Need Help

I want to create virtual env to run python udf test in it, but seems the script doesn't work. 

Do you have some thoughts?

chubei: Looks fine to me. Don't know why it doesn't work.

xudong963: Yeah, also works in my local

mediuminvader: are each action steps executed in the same shell? I mean if the env is activated at "Create python virtual env" step, will be still active at "Run actions-rs/cargo@v1" step?

chubei: Ah I think @mediuminvader got it here. Each step probablys runs in a different shell.

xudong963: Let me think how to fix it later :)

xudong963: done 😎

snork-alt (approved):

xudong963:
Let's merge and iterate later

PR #929, opened 2023-02-17
chore: Add `Source::can_start_from` with placeholder implementation

chubei:
Add `BuilderDag` which checks if sources can start from last checkpoint.

This is part of #928


chubei:
Pushed some more style changes to this PR:

- Remove `SourceFactory::prepare` and `ProcessorFactory::prepare` because they're not needed.
- Remove `DagSchemas::prepare` and call `SinkFactory::prepare` in `DagSchemas::new`, resulting in `prepare` being called multiple times. As a consequence:
  - Sql test sink factory's `prepare` implementation is moved to `build`
- Remove hardcoded `generated` folder. Generated proto files are under `api` now instead of `api/generated`.
- Remove `Sink::init` because it's not needed.

snork-alt (approved):

PR #931, opened 2023-02-17
Showcase usage of workspace dependencies

auterium:
I recently saw some dependency getting the version updated across multiple sub-crates which reminded me of this recent feature for `workspace.dependencies`. I've done the change for a few to show the functionality and if it makes sense, either I can move more or someone else can take the task :)

This feature allows for defining the version/features of a given dependency at the workspace's `Cargo.toml` so that any version change can be done in a single place rather than on every sub-crate


chubei:
@v3g42 Can you check this?

v3g42:
Hey @auterium. We have taken the approach of sharing common dependencies through `dozer-types` and be re-exported.  
https://github.com/getdozer/dozer/blob/main/dozer-types/src/lib.rs

That should be sufficient for the time being I believe. 

Workspace dependencies are nice too. But it makes sense to explore this later slightly down the line. Currently project is seeing a few refactors. It would cause merge conflicts at this stage.

auterium:
No worries @v3g42 makes sense.

PR #943, opened 2023-02-20
chore: Update connectors readme table layout

karolisg:


chloeminkyung (approved):

v3g42:
@karolisg @chloeminkyung I think it would be useful if there is some indication of the lifecycle for each of the methods.

When they get executed and what their purpose is in the overall implementation.

karolisg:
Added short description about methods usage in dozer commands

mediuminvader (approved):

PR #946, opened 2023-02-20
feat: Expose Pipeline config

mediuminvader:


dozer-orchestrator/src/pipeline/source_builder.rs
```diff
@@ -149,6 +149,11 @@ mod tests {
             endpoints: vec![],
             sql: None,
             home_dir: "test".to_string(),
+            cache_max_map_size: Default::default(),
```
chubei, line 152: These defaults are `0` right? Is this what we want?

dozer-types/src/models/app_config.rs
```diff
@@ -36,15 +40,77 @@ pub struct Config {
...
+    )]
+    pub cache_max_map_size: Option<u64>,
+
+    /// Pipeline lmdb max map size
+    #[prost(uint64, optional, tag = "11")]
+    #[serde(
+        default = "default_pipeline_max_map_size",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub pipeline_max_map_size: Option<u64>,
```
chubei, line 63: I can't find how this config is passed to the pipeline. Am I missing something?

dozer-orchestrator/src/pipeline/source_builder.rs
mediuminvader: nope, I forgot to use the default config functions, thanks for note this out...
should be fixed now

dozer-types/src/models/app_config.rs
mediuminvader: It's not not passed yet

dozer-orchestrator/src/utils.rs
```diff
@@ -15,6 +18,46 @@ pub fn get_cache_dir(config: &Config) -> PathBuf {
     AsRef::<Path>::as_ref(&config.home_dir).join("cache")
 }
 
+pub fn get_cache_max_map_size(config: &Config) -> usize {
```
chubei, line 21: Should we import and use the `default_*` functions here?

dozer-orchestrator/src/simple/orchestrator.rs
```diff
@@ -45,17 +47,27 @@ use tokio::sync::{broadcast, oneshot};
 pub struct SimpleOrchestrator {
     pub config: Config,
     pub cache_manager_options: CacheManagerOptions,
+    pub executor_options: ExecutorOptions,
```
chubei, line 50: Looks like calculated data. Maybe calculate it when needed instead of adding a member?

chubei (approved):
LGTM. Feel free to resolve the two comments and merge.

dozer-orchestrator/src/simple/orchestrator.rs
mediuminvader: Can you give me more details on this comment?
These parameters are coming from the configuration file

chubei: Exactly. These parameters are coming from the configuration file so I prefer not storing it as a field of the struct, and creating a method `fn get_executor_options(config: &Config) -> ExecutorOptions` instead. But that's just me and I'm not sure if this approach makes sense to you.

dozer-orchestrator/src/simple/orchestrator.rs
```diff
@@ -185,6 +187,7 @@ impl Orchestrator for SimpleOrchestrator {
             Some(sender),
             self.cache_manager_options.clone(),
             settings,
+            get_executor_options(&self.config),
```
mediuminvader, line 190: @chubei like this?

chubei: Exactly

PR #947, opened 2023-02-20
perf: Remove `RwCache::insert_schema` so we don't read the db on every insert

chubei:
# Core change

`RwCache::insert_schema` is removed, and `CacheManager::create_cache` not takes a `schemas` parameter.

All schema writes are performed at cache creation. Implemented in `LmdbRwCache::create`.

Schema lookup is changed from database read from hash map lookup. Implemented in `dozer-cache/src/cache/lmdb/cache/schema_database.rs`.

# Minor change

`RoCache::get_schema` and `RoCache::query` now returns `&Schema` to avoid cloning or repetitive lookup.

# Style change

Remove `ApiHelper` and use free functions instead. Functions only used in REST are moved to `dozer-api/src/rest/api_generator.rs`.

`MultiIndexFetch` is moved from `CacheError` to `ApiError`.

`OpenApiGenerator`, `ProtoGeneratorImpl` and `LmdbQueryHandler` now own references instead of values.

### DIFF
--- a/dozer-api/src/api_helper.rs
+++ b/dozer-api/src/api_helper.rs
@@ -1,241 +1,59 @@
 use crate::auth::Access;
 use crate::errors::{ApiError, AuthError};
+use dozer_cache::cache::expression::QueryExpression;
 use dozer_cache::cache::RecordWithId;
-use dozer_cache::cache::{expression::QueryExpression, index};
-use dozer_cache::errors::CacheError;
 use dozer_cache::{AccessFilter, CacheReader};
-use dozer_types::chrono::SecondsFormat;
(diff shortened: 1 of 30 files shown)


chubei:
@snork-alt How do I check whether performance gets improved?

snork-alt (approved):

chubei:
A very rough performance comparison.

## Before this commit

```
▸▹▹▹▹ trips_cache: Count: 272028, 3,127.4724/s
```

## After this commit

```
trips_cache: Count: 272270, 4,003.1985/s
```

Processing speed increases by 28%.

PR #961, opened 2023-02-21
feat: set operator lmdb support

chloeminkyung:


dozer-sql/src/pipeline/builder.rs
```diff
@@ -200,7 +200,7 @@ fn query_to_pipeline(
                     pipeline_idx,
                 )?;
             }
-            _ => panic!("{:?} is not supported", op),
+            _ => panic!("{op:?} is not supported"),
```
mediuminvader, line 203: Change this to an error

dozer-sql/src/pipeline/product/set.rs
```diff
@@ -45,14 +48,69 @@ impl SetOperation {
...
+        let write_txn = &mut txn.write();
+        match action {
+            SetAction::Insert => {
+                self.update_set_db(&lookup_key, 1, false, write_txn, *database);
+            }
+            SetAction::Delete => {
+                self.update_set_db(&lookup_key, 1, true, write_txn, *database);
+            }
+        }
+        let curr_count = self.get_occurrence(&lookup_key, write_txn, *database)?;
```
mediuminvader, line 64: can't we use the `new_count` value that has just been written with `update_set_db`, rather then reading it with `get_occurrence`?

dozer-sql/src/pipeline/product/set.rs
```diff
@@ -45,14 +48,69 @@ impl SetOperation {
         &self,
         action: SetAction,
```
mediuminvader, line 49: not sure we need this `action`, since we're returning the same value

dozer-sql/src/pipeline/builder.rs
chloeminkyung: sure

dozer-sql/src/pipeline/product/set.rs
chloeminkyung: ok - this was for double confirmation that we have recorded correctly

dozer-sql/src/pipeline/product/set.rs
chloeminkyung: we are using that actions in the `execute_union()` function -> https://github.com/getdozer/dozer/pull/961/files#diff-d84fd3c1d6ca93290e4336fcdfad00967e63bf4abc47280583974d5af8815d56R57

PR #963, opened 2023-02-21
feat: Include id in grpc event stream

chubei:
Currently only `INSERT` operation gets `id`.

typed:

Screenshot 2023-02-21 at 4 18 45 PM

common:

Screenshot 2023-02-21 at 4 20 54 PM

### DIFF
--- a/dozer-api/protos/films.proto
+++ b/dozer-api/protos/films.proto
@@ -73,7 +73,9 @@ message FilmEvent {
   // Old record data, only applicable for UPDATE type.
   optional Film old = 2;
   // New record data.
-  optional Film new = 3;
+  Film new = 3;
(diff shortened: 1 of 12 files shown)


v3g42:
Just to make it clear, do you wanna give it a dozer prefix ? Like __dozer_row_id or something?

Also wonder if there is a reason you called it new_id specifically?

chubei:
> Just to make it clear, do you wanna give it a dozer prefix ? Like __dozer_row_id or something?
> 
> Also wonder if there is a reason you called it new_id specifically?

It's outside the record, so don't need a special prefix. I call it `new_id` because the corresponding record is called `new`.

v3g42 (approved):

PR #965, opened 2023-02-21
fix: discord notifications

v3g42:
- [x] Issues / Discussions on Discord ideas channel
- [x] Github pushes to Github channel
- [x] releases to releases


chubei:
Looks like it still has the same old problem. Forks cannot use repo secrets.

v3g42:
> Looks like it still has the same old problem. Forks cannot use repo secrets.

Only for the file `pulls`.

v3g42:
> > > Looks like it still has the same old problem. Forks cannot use repo secrets.
> > 
> > 
> > Only for the file `pulls`.
> 
> So `Dozer Pulls` will fail? That's a problem. And `Dozer CI` is failing because it runs on `pull_request`.

Yes looking into it right now. Worst case, we can disable pulls.yaml file.

PR #982, opened 2023-02-21
chore: introduce python feature flag

v3g42:
Running into issues running because of new dependencies.
```
bs"
  = note: ld: warning: directory not found for option '-L/opt/homebrew/opt/odbc/lib'
          ld: warning: directory not found for option '-L/opt/homebrew/opt/odbc/lib'
          Undefined symbols for architecture arm64:
            "_ERR_get_error_all", referenced from:
                openssl::error::Error::get::h10687d4ca56e2d56 in libopenssl-2efd15f3d78a2746.rlib(openssl-2efd15f3d78a2746.openssl.d451f57b-cgu.5.rcgu.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Moving python under a feature flag `python`. 

@xudong963 Can you verify ?

### DIFF
--- a/dozer-sql/src/pipeline/expression/execution.rs
+++ b/dozer-sql/src/pipeline/expression/execution.rs
@@ -54,6 +54,7 @@ pub enum Expression {
         pattern: Box<Expression>,
         escape: Option<char>,
     },
+    #[cfg(feature = "python")]
     PythonUDF {
         name: String,
         args: Vec<Expression>,
@@ -102,6 +103,7 @@ impl Expression {
                         .as_str()
                     + ")"
             }
+            #[cfg(feature = "python")]
             Expression::PythonUDF { name, args, .. } => {
                 name.to_string()
                     + "("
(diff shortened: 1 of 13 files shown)


dozer-sql/src/pipeline/expression/execution.rs
```diff
@@ -205,12 +204,23 @@ impl ExpressionExecutor for Expression {
                 right,
             } => operator.evaluate(schema, left, right, record),
             Expression::ScalarFunction { fun, args } => fun.evaluate(schema, args, record),
+
+            #[allow(unused_variables)]
             Expression::PythonUDF {
```
auterium, line 209: Would it be more readable to feature-flag the `Expression::PythonUDF` variant instead? That way if a user tries to use a Python UDF when the feature is not enabled, then it would return a proper error instead of killing the thread (through the `panic!()` in line 222)

v3g42: Makes sense. Changed. Thanks @auterium

mediuminvader (approved):

xudong963:
I tested the ticket in my local.
image

I think the python_udf test doesn't run, I'm finding reason.

v3g42:
> I tested the ticket in my local. image
> 
> I think the python_udf test doesn't run, I'm finding reason.

You will have to run with `python` feature enabled. 
```
--features=python
```

Issue #960, opened 2023-02-21
Implement `Field::Point` to PyObject


xudong963:
Leave it as todo, after the ticket merged, I'll open a ticket to complete it.

_Originally posted by @xudong963 in https://github.com/getdozer/dozer/pull/912#discussion_r1112586082_

PR #984, opened 2023-02-21
feat: enhance python udf 

xudong963:
1. https://github.com/getdozer/dozer/issues/960
2. refactor test framework

### DIFF
--- a/dozer-sql/src/pipeline/expression/python_udf.rs
+++ b/dozer-sql/src/pipeline/expression/python_udf.rs
@@ -15,7 +15,7 @@ pub fn evaluate_py_udf(
     schema: &Schema,
     name: &str,
     args: &[Expression],
-    return_type: &FieldType,
+    return_type: &Option<FieldType>,
     record: &Record,
 ) -> Result<Field, PipelineError> {
     let values = args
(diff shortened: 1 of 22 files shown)


dozer-sql/src/pipeline/expression/python_udf.rs
```diff
@@ -15,7 +15,7 @@ pub fn evaluate_py_udf(
     schema: &Schema,
     name: &str,
     args: &[Expression],
-    return_type: &FieldType,
+    return_type: &Option<FieldType>,
```
mediuminvader, line 18: What's the purpose of Option here, is this the return type of the UDF?
Do we allow to don't specify the return type? how it will works?

dozer-sql/src/pipeline/expression/python_udf.rs
```diff
@@ -45,18 +45,21 @@ pub fn evaluate_py_udf(
 
         let args = PyTuple::new(py, values);
         let res = function.call1(args)?;
-        Ok(match return_type {
+        if return_type.is_none() {
```
mediuminvader, line 48: Is this the case the return type for python udf is not specified?

xudong963: Yes

dozer-sql/src/pipeline/expression/python_udf.rs
xudong963: Yes, we support python udf has no return value. Such as:

```python
def hello():
    print("hello")
```

xudong963:
Any new comments?

mediuminvader (approved):

PR #994, opened 2023-02-22
feat: Implement start from in postgres connector

### LINKED ISSUE #990: Connector interface adjustment with unit test cases
The purpose of this change is to implement support for `can_start_from` method in connector. It is used in https://github.com/getdozer/dozer/pull/929/files#diff-1683429d0a37884e20572963875a8f4dca9b5d96d3161b4a8a5eaa1a32b75b6bR186

- [x] `postgres` can resume if the following conditions are met:
  - [x] Replication slot exists
  - [x] Tables already are added to publication slot
  - [x] Compare `confirmed_flush_lsn` with lsn from method arguments. `confirmed_flush_lsn` should be before 
  - [x] <strike>LSN is bigger than 0 (0 LSN is used for snapshotting phase)</strike> _Connector will not receive lsn if snapshotting is not done_
- [ ] `object_store` always true for append-only ingestion
- [ ] `snowflake`:
  - [ ]  stream already exist

karolisg:
Closes #990

### DIFF
--- a/dozer-ingestion/src/connectors/postgres/iterator.rs
+++ b/dozer-ingestion/src/connectors/postgres/iterator.rs
@@ -163,7 +163,7 @@ impl<'a> PostgresIteratorHandler<'a> {
                 ingestor: self.ingestor,
                 connector_id: self.connector_id,
             };
-            tables = snapshotter.sync_tables(details.tables.clone(), self.lsn.borrow().as_ref())?;
+            tables = snapshotter.sync_tables(details.tables.clone())?;
 
             debug!("\nInitialized with tables: {:?}", tables);
 
--- a/dozer-ingestion/src/connectors/postgres/iterator.rs
+++ b/dozer-ingestion/src/connectors/postgres/iterator.rs
@@ -10,16 +10,14 @@ use std::str::FromStr;
 use std::sync::Arc;
 
 use crate::connectors::postgres::connection::helper;
+use crate::connectors::postgres::replication_slot_helper::ReplicationSlotHelper;
 use crate::connectors::postgres::replicator::CDCHandler;
 use crate::connectors::postgres::snapshotter::PostgresSnapshotter;
-use crate::errors::ConnectorError::UnexpectedQueryMessageError;
 use crate::errors::PostgresConnectorError::{
-    LSNNotStoredError, LsnNotReturnedFromReplicationSlot, LsnParseError,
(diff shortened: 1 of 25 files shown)


dozer-ingestion/src/connectors/postgres/iterator.rs
```diff
@@ -161,7 +164,12 @@ impl<'a> PostgresIteratorHandler<'a> {
                 ingestor: self.ingestor,
                 connector_id: self.connector_id,
             };
-            tables = snapshotter.sync_tables(details.tables.clone(), self.lsn.borrow().as_ref())?;
+            tables = snapshotter.sync_tables(details.tables.clone())?;
+
+            let lsn = self.lsn.borrow().map_or(0, |(lsn, _)| u64::from(lsn));
+            self.ingestor
+                .handle_message(((lsn, 0), SnapshottingDone))
```
chubei, line 171: Is lsn guarenteed to be positive?

dozer-ingestion/src/ingestion/ingestor.rs
```diff
@@ -80,6 +80,7 @@ impl Ingestor {
             }
             IngestionMessage::Commit(_event) => {}
             IngestionMessage::Begin() => {}
+            IngestionMessage::SnapshottingDone => {}
```
chubei, line 83: Just making a mark here so later I know where to start passing along this message

chubei (approved):
Great!

dozer-ingestion/src/connectors/postgres/iterator.rs
karolisg: Yes, it is coming from a slot creation query. LSN struct is `pub struct PgLsn(u64);`.

chubei: And it's never `0`?

karolisg: Yes. LSN can be between 0/0 and FFFFFFFF/FFFFFFFF. So theoretically it can be 0, but before creating slot, we have already created publication, so we can assume that slot creation message cannot have LSN 0/0.

chubei: Wonderful. Maybe put a comment line there so later we don't get confused again and believe this `0` is a potential bug?

PR #1008, opened 2023-02-23
fix: Invalid column error

mediuminvader:


chubei (approved):
## Before

`ERROR [pipeline] Validation error: internal error: Invalid expression: .trip_distance`

## After

`ERROR [pipeline] Validation error: internal error: SQL Error: Invalid column name .trip_distance.`

Looks good to me!

Issue #1012, opened 2023-02-23
Field::TimeStamp is not working with gRPC. 


v3g42:
### Error
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidType { field: FieldDescriptor { name: "last_update", full_name: "dozer.generated.actor.Actor.last_update", json_name: "lastUpdate", number: 4, kind: google.protobuf.Timestamp, cardinality: Optional, containing_oneof: None, default_value: None, is_group: false, is_list: false, is_map: false, is_packed: false, supports_presence: true }, value: String("2023-01-22T16:34:58.253Z") }', /Users/vivek/.cargo/registry/src/github.com-1ecc6299db9ec823/prost-reflect-0.10.2/src/dynamic/mod.rs:213:47
```

### Config
```
app_name: first_app

# Define your connections here
connections:
  - config: !Postgres
      user: postgres
      password: postgres
      host: 127.0.0.1
      port: 5433
      database: pagila
    name: postgres_connection

# Each table is a source 
sources:
  - name: actor
    table_name: actor
    columns: []
    connection: !Ref postgres_connection
    refresh_config: null

# Each of the endpoints are configured here
endpoints:
  - name: actor
    table_name: actor
    path: /api/actor
    index: null

```

chloeminkyung:
This [line](https://github.com/getdozer/dozer/blob/main/dozer-api/src/grpc/typed/helper.rs#L74) of code for protobuf types are going to cause similar issue with `PointValue`, `DecimalValue`, `DateValue`.

Like I did to fix the issue with timestamp,

```
GrpcTypes::value::Value::TimestampValue(ts) => Value::Message(ts.transcode_to_dynamic()),
```

Custom types need to be wrapped or mapped to matching type with the `field_descriptor`'s `kind`

mediuminvader:
Shall we create issues for those 3 types as well?

chloeminkyung:
> Shall we create issues for those 3 types as well?
it depends, currently we are pushing `date` as string back to dynamic grpc types but not sure how we are going to handle the `point` and `decimal`

chloeminkyung:
All of the `Point`, `Decimal`, `Timestamp` support has been covered in the same PR

PR #1015, opened 2023-02-23
feat: `decimal` and `binary` support for grpc

chloeminkyung:
feat: decimal and binary support for grpc


v3g42:
@chloeminkyung While you are at it, I ran into this today. can you check this too? 
https://github.com/getdozer/dozer/issues/1012

chloeminkyung:
> @chloeminkyung While you are at it, I ran into this today. can you check this too? #1012

will address that ticket is separate pr

PR #1035, opened 2023-02-25
fix: add postgres columns validations and connector test

hoangnh93:


dozer-ingestion/src/connectors/postgres/connection/validator.rs
```diff
@@ -176,17 +177,50 @@ fn validate_tables(
...
+            .iter()
+            .find(|x| x.table_name == table_name)
+            .unwrap()
+            .clone()
+            .columns;
+
+        if let Some(column_info) = columns {
+            let column_result = client
+                .query(
+                    "SELECT column_name FROM information_schema.columns WHERE table_name = ($1)",
```
karolisg, line 195: It doesn't look efficient to run query on every table separately

karolisg (changes requested):

dozer-ingestion/src/connectors/postgres/snapshotter.rs
```diff
@@ -112,3 +112,344 @@ impl<'a> PostgresSnapshotter<'a> {
...
+            let (ingestor, mut _iterator) = Ingestor::initialize_channel(ingestion_config);
+
+            let snapshotter = PostgresSnapshotter {
+                tables: Some(tables.clone()),
+                conn_config: config,
+                ingestor: &ingestor,
+                connector_id: connector.id,
+            };
+
+            let actual = snapshotter.sync_tables(Some(input_tables));
```
karolisg, line 213: `sync_tables` no longer return tables. In this test, you need to verify what messages ingestor gets.

karolisg (changes requested):

dozer-ingestion/src/connectors/postgres/connection/validator.rs
hoangnh93: updated!

dozer-ingestion/src/connectors/postgres/snapshotter.rs
hoangnh93: updated!

karolisg (approved):

PR #1060, opened 2023-02-26
fix: `timestamp` support for grpc

chloeminkyung:
fix: `timestamp` support for grpc

### DIFF
--- a/dozer-api/src/grpc/typed/helper.rs
+++ b/dozer-api/src/grpc/typed/helper.rs
@@ -5,7 +5,7 @@ use crate::generator::protoc::generator::{
 use crate::grpc::types_helper::map_record;
 use dozer_cache::cache::RecordWithId;
 use dozer_types::grpc_types::types as GrpcTypes;
-use prost_reflect::{DynamicMessage, Value};
+use prost_reflect::{DynamicMessage, ReflectMessage, Value};
 
 use super::TypedResponse;
 
@@ -70,8 +70,8 @@ fn interval_value_to_pb(value: GrpcTypes::Value) -> Option<prost_reflect::Value>
             Value::Bytes(prost_reflect::bytes::Bytes::from(n))
         }
         GrpcTypes::value::Value::PointValue(_p) => todo!(),
-        GrpcTypes::value::Value::DecimalValue(_d) => todo!(),
-        GrpcTypes::value::Value::TimestampValue(_t) => todo!(),
+        GrpcTypes::value::Value::DecimalValue(d) => todo!(),
+        GrpcTypes::value::Value::TimestampValue(ts) => Value::Message(ts.transcode_to_dynamic()),
         GrpcTypes::value::Value::DateValue(_) => todo!(),
     })
 }
--- a/dozer-api/src/grpc/typed/helper.rs
+++ b/dozer-api/src/grpc/typed/helper.rs
@@ -70,7 +70,7 @@ fn interval_value_to_pb(value: GrpcTypes::Value) -> Option<prost_reflect::Value>
             Value::Bytes(prost_reflect::bytes::Bytes::from(n))
         }
         GrpcTypes::value::Value::PointValue(_p) => todo!(),
-        GrpcTypes::value::Value::DecimalValue(d) => todo!(),
+        GrpcTypes::value::Value::DecimalValue(_d) => todo!(),
         GrpcTypes::value::Value::TimestampValue(ts) => Value::Message(ts.transcode_to_dynamic()),
         GrpcTypes::value::Value::DateValue(_) => todo!(),
     })
--- a/dozer-api/src/grpc/types_helper.rs
+++ b/dozer-api/src/grpc/types_helper.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+use prost_reflect::prost_types::Timestamp;
 use dozer_cache::cache::RecordWithId as CacheRecordWithId;
 use dozer_types::chrono::SecondsFormat;
 use dozer_types::ordered_float::OrderedFloat;
@@ -108,8 +110,8 @@ fn field_to_prost_value(f: Field) -> Value {
         },
         Field::Decimal(d) => map_decimal(d),
         Field::Timestamp(ts) => Value {
-            value: Some(value::Value::StringValue(
-                ts.to_rfc3339_opts(SecondsFormat::Millis, true),
+            value: Some(value::Value::TimestampValue(
+                Timestamp::from_str(&ts.to_rfc3339_opts(SecondsFormat::Millis, true)).unwrap_or_default(),
             )),
         },
         Field::Bson(b) => Value {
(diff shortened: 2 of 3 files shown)


chloeminkyung:
failing coverage cause of following error,
```
/usr/bin/ld: final link failed: No space left on device
          collect2: error: ld returned 1 exit status
```

@chubei , is this the cache issue that you mention last time?

chubei:
> failing coverage cause of following [error](https://github.com/getdozer/dozer/actions/runs/4275567039/jobs/7443016519#step:10:28),
> 
> ```
> /usr/bin/ld: final link failed: No space left on device
>           collect2: error: ld returned 1 exit status
> ```
> 
> @chubei , is this the cache issue that you mention last time?

Maybe... clear the cache and retry?

dozer-api/src/grpc/types_helper.rs
```diff
@@ -108,8 +110,9 @@ fn field_to_prost_value(f: Field) -> Value {
         },
         Field::Decimal(d) => map_decimal(d),
         Field::Timestamp(ts) => Value {
-            value: Some(value::Value::StringValue(
-                ts.to_rfc3339_opts(SecondsFormat::Millis, true),
+            value: Some(value::Value::TimestampValue(
+                Timestamp::from_str(&ts.to_rfc3339_opts(SecondsFormat::Millis, true))
```
chubei, line 114: Do we have to go through `String`?

dozer-types/protos/types.proto
```diff
@@ -101,18 +101,15 @@ message RustDecimal {
...
-
-    PointType point_value = 12; // Point type
+    uint64 uint_value = 1;     // Unsigned 64 bit integer.
+    int64 int_value = 2;       // Signed 64 bit integer.
+    double float_value = 3;    // 32 bit floating point number.
+    bool bool_value = 4;       // Boolean.
+    string string_value = 5;   // UTF-8 string.
+    bytes bytes_value = 7;     // Binary data.
+    RustDecimal decimal_value = 8;                 // Decimal value.
+    google.protobuf.Timestamp timestamp_value = 9; // DateTime & Timestamp.
```
chubei, line 111: `Date` is still converted to `String`

chloeminkyung: yes - because proto authors also want `Date` to be in64 or `string` no native type to be mapped into, however, timestamp is available for us to use as return value type for our typed fields

dozer-api/src/grpc/types_helper.rs
chloeminkyung: in case we want to return this as `google.protobuf.Timestamp` then we don't need to make this as String

chloeminkyung: I get where your comment is coming from, but we need to decide whether we want to keep the result of the fields to be

currently, with this pr's change below is the output from typed grpc

```
{
            "typ": "String",
            "name": "create_date",
            "nullable": false
        },
        {
            "typ": "Timestamp",
            "name": "last_update",
            "nullable": true
        },
```

chubei: > decide whether we want to keep the result of the fields to be

This is the question I wanted to ask in the other comment.

This comment is because I see when we convert a `Field::Timestamp` to `grpc::Timestamp`, it goes like this `Filed::Timpstamp` -> `String` -> `grpc::Timestamp`. I was wondering if this is necessary. Can we save the formatting and parsing cost?

dozer-api/src/generator/protoc/generator/implementation.rs
```diff
@@ -167,15 +167,32 @@ impl<'a> ProtoGeneratorImpl<'a> {
...
-            Ok(RecordDesc {
-                message,
-                version_field,
-            })
-        }
+        let record_desc_from_message =
+            |message: MessageDescriptor| -> Result<RecordDesc, GenerationError> {
+                let version_field = get_field(&message, "__dozer_record_version")?;
+                let point_values = descriptor
+                    .get_message_by_name("dozer.types.PointType")
```
chubei, line 174: `dozer.types.PointType` and `dozer.types.RustDecimal` are hardcoded in two different places. Better create constants for them.

It's better if `unwrap` can be replaced by an error with proper context.

dozer-api/src/grpc/typed/helper.rs
```diff
@@ -59,7 +59,11 @@ fn internal_record_to_pb(record: GrpcTypes::Record, record_desc: &RecordDesc) ->
     msg
 }
 
-fn interval_value_to_pb(value: GrpcTypes::Value) -> Option<prost_reflect::Value> {
+fn interval_value_to_pb(
+    value: GrpcTypes::Value,
+    _field: &FieldDescriptor,
```
chubei, line 64: Unused parameter is better removed.

dozer-api/src/grpc/typed/helper.rs
```diff
@@ -69,10 +73,30 @@ fn interval_value_to_pb(value: GrpcTypes::Value) -> Option<prost_reflect::Value>
...
             Value::Bytes(prost_reflect::bytes::Bytes::from(n))
         }
-        GrpcTypes::value::Value::PointValue(_p) => todo!(),
-        GrpcTypes::value::Value::DecimalValue(_d) => todo!(),
-        GrpcTypes::value::Value::TimestampValue(_t) => todo!(),
-        GrpcTypes::value::Value::DateValue(_) => todo!(),
+        GrpcTypes::value::Value::PointValue(p) => {
+            let point_type_desc = descriptor.clone().point_field.message;
+            let x_field_desc = descriptor.clone().point_field.x;
+            let y_field_desc = descriptor.clone().point_field.y;
```
chubei, line 79:
```suggestion
            let point_type_desc = descriptor.point_field.message.clone();
            let x_field_desc = &descriptor.point_field.x;
            let y_field_desc = &descriptor.point_field.y;
```

Same below.

dozer-api/src/grpc/types_helper.rs
```diff
@@ -108,9 +108,10 @@ fn field_to_prost_value(f: Field) -> Value {
         },
         Field::Decimal(d) => map_decimal(d),
         Field::Timestamp(ts) => Value {
-            value: Some(value::Value::StringValue(
-                ts.to_rfc3339_opts(SecondsFormat::Millis, true),
-            )),
+            value: Some(value::Value::TimestampValue(Timestamp {
+                seconds: ts.timestamp(),
+                nanos: (ts.timestamp_nanos() - ts.timestamp() * 1000000000) as i32,
```
chubei, line 113:
```suggestion
                nanos: ts.timestamp_subsec_nanos() as i32,
```

chubei (approved):
Looks good to me! Feel free to resolve the comments without change and merge.

dozer-api/src/generator/protoc/generator/implementation.rs
chloeminkyung: alright

dozer-api/src/grpc/typed/helper.rs
chloeminkyung: yup!

dozer-api/src/grpc/typed/helper.rs
chloeminkyung: ok!

dozer-api/src/grpc/types_helper.rs
chloeminkyung: 🫡

PR #1061, opened 2023-02-26
feat: Blue green cache

### LINKED ISSUE #928: Blue/Green cache

chubei:
Closes #928

Removed `clean` on dozer startup.

Relavant logs:

```
...
2023-02-26T18:38:25.832862Z  INFO [pipeline] Inconsistent, resetting    
2023-02-26T18:38:25.841439Z  INFO [pipeline] Cache stocks writing to 82ffe650-9e40-4dea-b6d4-44503123994e while serving 6e7f4187-1e15-4214-ba79-300c0172c648    
2023-02-26T18:38:25.848192Z  INFO [pipeline] Cache stocks_meta writing to 542af2db-2096-4ef7-94d6-807592ed7077 while serving 04cd4326-4e5e-4837-8029-255660bbff07    
...  
2023-02-26T18:38:27.098957Z  INFO [api] Redirecting cache stocks to 82ffe650-9e40-4dea-b6d4-44503123994e    
2023-02-26T18:38:27.428381Z  INFO [api] Redirecting cache stocks_meta to 542af2db-2096-4ef7-94d6-807592ed7077    
```

### DIFF
--- a/dozer-api/src/errors.rs
+++ b/dozer-api/src/errors.rs
@@ -54,10 +54,11 @@ impl ApiError {
 }
 
 #[derive(Error, Debug)]
-pub enum GRPCError {
-    #[error("Internal GRPC server error: {0}")]
+pub enum GrpcError {
+    #[error("Internal gRPC server error: {0}")]
     InternalError(#[from] BoxedError),
-
+    #[error("Cannot send to broadcast channel")]
+    CannotSendToBroadcastChannel,
     #[error(transparent)]
     SerizalizeError(#[from] serde_json::Error),
     #[error("Missing primary key to query by id: {0}")]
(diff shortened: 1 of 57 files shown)


chubei:
There're still two problems:

1. The processor nodes don't have consistent names across dozer runs, so the pipeline is never consistent and always starts the sources from scratch.
2. Currently cache switch happens on the first `SnapshottingDone` message from whichever ancestor sources. I'm not sure if this is a good strategy.

v3g42 (approved):

v3g42:
> There're still two problems:
> 
> 1. The processor nodes don't have consistent names across dozer runs, so the pipeline is never consistent and always starts the sources from scratch.
> 2. Currently cache switch happens on the first `SnapshottingDone` message from whichever ancestor sources. I'm not sure if this is a good strategy.

I don't fully follow the implications of first `SnapshottingDone` implications. Ideally it should be based on when the versions in the cache caught up with the older versions. I think Debezium describes this in full detail in some article.

Lets discuss this separately. But for now I think it is fair enough.

Also, in some cases users might want to explicitly trigger the switch.

chubei:
> Also, in some cases users might want to explicitly trigger the switch.

We can add a method in the internal grpc server, creating the new alias and notifying API server.

PR #1070, opened 2023-02-27
feature: Add schema support to postgres connector

karolisg:


v3g42:
@karolisg Ideally i think we should be able to query multiple schemas using the same configuration. Is that not possible. ?

This implements restricts me to a single schema even though connection can be reused.

karolisg:
> @karolisg Ideally i think we should be able to query multiple schemas using the same configuration. Is that not possible. ?
> 
> This implements restricts me to a single schema even though connection can be reused.

Yes, that is possible. Maybe it is a good idea to add `schema` to `TableInfo` because it can be used in most connectors.

karolisg:
I will continue work on this after https://github.com/getdozer/dozer/pull/1075 is merged, because otherwise I will need solve conflicts in schema helper

mediuminvader (approved):

PR #1075, opened 2023-02-27
fix: Fix check of supported tables in postgres

### LINKED ISSUE #1072: Postgres `get_schemas` throws a runtime error for views.
Postgres breaks during `get_schemas` when source db has `views`. 

The following line breaks as this value is obtained as `NULL`
https://github.com/getdozer/dozer/blob/main/dozer-ingestion/src/connectors/postgres/schema_helper.rs#L295

karolisg:
closes #1072

### DIFF
--- a/dozer-ingestion/src/connectors/mod.rs
+++ b/dozer-ingestion/src/connectors/mod.rs
@@ -85,11 +85,17 @@ pub struct TableInfo {
     pub columns: Option<Vec<ColumnInfo>>,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash)]
 #[serde(crate = "self::serde")]
 pub struct ColumnInfo {
     pub name: String,
     pub data_type: Option<String>,
+}
+
+impl ColumnInfo {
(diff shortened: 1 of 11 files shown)


chubei (commented):
Don't really understand this code

karolisg:
> Don't really understand this code

This PR contains several changes:
1) Validation of table type added to schema helper
2) Did some minor refactorings in schema helper to reduce complex and not clear type definitions
3) Add table type validator 
4) Moved schema tests to separate file
5) Commented kafka tests, because they are not used at the moment

chubei:
> This PR contains several changes:
> 
> 1. Validation of table type added to schema helper
> 2. Did some minor refactorings in schema helper to reduce complex and not clear type definitions
> 3. Add table type validator
> 4. Moved schema tests to separate file
> 5. Commented kafka tests, because they are not used at the moment

Thanks for the explaination! I kind of get what this PR is trying to do. Just don't have enough postgres knowledge to review it.

v3g42:
@karolisg the query takes very long to run. Is this the only way to obtain metadata in SQL

karolisg:
> @karolisg the query takes very long to run. Is this the only way to obtain metadata in SQL

I will solve it in different PR. This one already has a lot changes.

chloeminkyung (approved):

PR #1079, opened 2023-02-27
feat: implement progress for sources

v3g42:


snork-alt:
we could use something liek this:
https://blog.logrocket.com/rust-and-tui-building-a-command-line-interface-in-rust/

chubei:
> we could use something liek this: https://blog.logrocket.com/rust-and-tui-building-a-command-line-interface-in-rust/

Yes a TUI would be great.

PR #1093, opened 2023-02-28
fix: timestamp diff, add, mul, div

chloeminkyung:
fix: timestamp diff, add, mul, div

- utilizing unix milliseoncds of the timestamp to perform +, -, *, /

todo: support for `Interval`


chubei (commented):
I'm not sure if this is what's in @mediuminvader's mind. I'll leave to him to decide.

PR #1094, opened 2023-02-28
feat: Read data from postgres snapshot in parallel

karolisg:


dozer-ingestion/src/connectors/postgres/snapshotter.rs
```diff
@@ -34,64 +40,96 @@ impl<'a> PostgresSnapshotter<'a> {
             .collect())
     }
 
-    pub fn sync_tables(&self, tables: Vec<TableInfo>) -> Result<Vec<TableInfo>, ConnectorError> {
+    pub fn sync_table(
+        table_info: TableInfo,
+        conn_config: tokio_postgres::Config,
+        sender: Sender<Option<Operation>>,
+    ) -> Result<(), ConnectorError> {
         let client_plain = Arc::new(RefCell::new(
```
chubei, line 48: `Arc` and `RefCell` can be removed?

dozer-ingestion/src/connectors/postgres/snapshotter.rs
```diff
@@ -34,64 +40,96 @@ impl<'a> PostgresSnapshotter<'a> {
             .collect())
     }
 
-    pub fn sync_tables(&self, tables: Vec<TableInfo>) -> Result<Vec<TableInfo>, ConnectorError> {
+    pub fn sync_table(
```
chubei, line 43: Errors in this function will not be reported? Is that a problem?

chubei (approved):

PR #1102, opened 2023-03-01
chore: Add authors to packages

karolisg:
Deb package generation requires authors defined in packages


chubei (approved):
How about just the core team...

PR #1112, opened 2023-03-02
fix: Sink should only be built after all source checkpoints are checked

chubei:


chubei:
The CI failure is bogus. It's the accidentatlly introduced `pull_request_target` trigger that failed. `pull_request` triggered workflow passes.

snork-alt (approved):

PR #1125, opened 2023-03-03
chore: Use `--llvm` to speed up coverage parsing

chubei:


chubei:
Old `12m12s`. New `11m55s`. Don't work.

PR #1135, opened 2023-03-05
refactor: pipeline memory storage

snork-alt:


dozer-types/src/types/mod.rs
```diff
@@ -228,6 +227,28 @@ impl Record {
...
+    pub fn get_key_fields(&self, schema: &Schema) -> Vec<Field> {
+        debug_assert!(
+            !&schema.primary_index.is_empty(),
+            "Primary key indexes cannot be empty"
+        );
+        let mut idx_values = Vec::with_capacity(schema.primary_index.len());
+        for i in &schema.primary_index {
+            idx_values.push(self.values[*i].clone());
+        }
+        idx_values
```
chubei, line 239:
```suggestion
        self.get_fields_by_indexes(&schema.primary_index)
```

chubei (commented):
Now that `RecordReader` is removed, can you explain the purpose of `RecordWriter`?

snork-alt:
`RecordWriter` is used to store the record when the port is stateful. Storing is necessary in some situations like Postgres where updates only send the full new record, but only the primary key of the old record. `RecordWriter` stores the records, and lookups the older version or the record to create a new Operation message. Maybe the name `RecordWriter` is not appropriate anymore. I will change it to something else

chubei (approved):
Merge?

PR #1173, opened 2023-03-08
use hash as a key in `RecordWriter` instead of `Vec<Field>`

snork-alt:


chubei (commented):
What if there's a hash collision?

snork-alt:
We will allow customizing the hash size to avoid collisions based on the volume of the data. I realized in many situations, hashing is the only option we have to keep track of data in memory. Same for `UNIONs` for instance: we cannot afford to keep in memory all data to check for duplicates. We use a bloom filter and allow fine-tuning if the user wants. Several analytical databases do that.

chubei (approved):

PR #1175, opened 2023-03-08
feat: Window processor

mediuminvader:
Implementing WINDOW function as a processor of the DAG

see https://github.com/getdozer/dozer/discussions/893 for details


snork-alt (approved):
When we return the schema in the factory do we add the `window_start` and `window-end` to the pk ?

PR #1176, opened 2023-03-08
feat: implement tracing using open telemetry

v3g42:


dozer-tracing/src/helper.rs
```diff
@@ -0,0 +1,116 @@
...
+    let fields = vec![
+        FieldDefinition {
+            name: "id".to_string(),
+            typ: FieldType::UInt,
+            nullable: false,
+            source: SourceDefinition::Dynamic,
+        },
+        FieldDefinition {
+            name: "trace_id".to_string(),
+            typ: FieldType::Binary,
```
chubei, line 52: We're going to change this to `UInt128` when we have it?

chubei (approved):

dozer-tracing/src/helper.rs
v3g42: yes

PR #1177, opened 2023-03-08
feat: array Field & FieldType

chloeminkyung:
feat: array Field & FieldType

### DIFF
--- a/dozer-types/src/types/field.rs
+++ b/dozer-types/src/types/field.rs
@@ -1,12 +1,11 @@
-use crate::errors::types::{DeserializationError, TypeError};
 #[allow(unused_imports)]
+use crate::errors::types::{DeserializationError, TypeError};
 use chrono::{DateTime, Datelike, FixedOffset, LocalResult, NaiveDate, TimeZone, Utc};
 use ordered_float::OrderedFloat;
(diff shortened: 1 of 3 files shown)


dozer-types/src/types/field.rs
```diff
@@ -42,6 +42,7 @@ pub enum FieldBorrow<'a> {
     Date(NaiveDate),
     Bson(&'a [u8]),
     Point(DozerPoint),
+    Array(Vec<Field>),
```
snork-alt, line 45: Should be `&'a [Field]`

v3g42:
@chloeminkyung I suggest doing the following changes as well in one PR. 
- Modify TimeStamp to contain units (Refer to this issue https://github.com/getdozer/dozer/issues/1181)
- Modify Date to contain Timezone
- Introduce Duration with Units

dozer-types/src/types/field.rs
chloeminkyung: it cannot be done that way because of #[derive(Deserialize)]

```
the trait bound `&'a [types::field::Field]: ingestion_types::_::_serde::Deserialize<'_>` is not satisfied [E0277] the trait `ingestion_types::_::_serde::Deserialize<'_>` is not implemented for `&'a [types::field::Field]`
```

PR #1194, opened 2023-03-09
chore: Run two full e2e test cases in PR runs

v3g42:
With this, we have two e2e examples running as a test part of PRs.

- GRPC Ingestor direct to SInk
- Object Source through SQL


dozer-tests/src/tests/e2e/basic_sql.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use dozer_api::actix_web::rt::Runtime;
+use dozer_orchestrator::{simple::SimpleOrchestrator as Dozer, Orchestrator};
+use dozer_types::grpc_types::common::{
+    common_grpc_service_client::CommonGrpcServiceClient, QueryRequest,
+};
+use dozer_types::{models::app_config::Config, serde_yaml};
+use tempdir::TempDir;
+fn get_config() -> Config {
+    let path = TempDir::new("tests").unwrap();
```
chubei, line 14: Does it work? I thought `tempdir` deletes the temporary directory on `TempDir::drop`.

chubei (approved):

dozer-tests/src/tests/e2e/basic_sql.rs
v3g42: Seems to work.

v3g42:
Currently this test can fail if `dozer` takes too long to start. I have included `1000` seconds before `query` happens. 
Otherwise we can increase the timeout.

Issue #956, opened 2023-02-21
Introduce Nested Record structure in dozer types


v3g42:
- Add Record as a Field Type. 
- (or) Array type

I am wondering if we can remove Bson altogether and just stick with Binary. 

```rust
pub const DATE_FORMAT: &str = "%Y-%m-%d";
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord)]
pub enum Field {
    UInt(u64),
    Int(i64),
    Float(OrderedFloat<f64>),
    Boolean(bool),
    String(String),
    Text(String),
    Binary(Vec<u8>),
    Decimal(Decimal),
    Timestamp(DateTime<FixedOffset>),
    Date(NaiveDate),
    Bson(Vec<u8>),
    Null,
}
```

chubei:
`Bson` is for postgres `JSONB` data type last time I asked @snork-alt

revskill10:
Do you have ETA for this feature ? I'm using JSONB and Array Type for our database and got error `__jsonb not supported` on array jsonb type.

chloeminkyung:
> Do you have ETA for this feature ? I'm using JSONB and Array Type for our database and got error `__jsonb not supported` on array jsonb type.

Hi @revskill10 , we have nested type support on its way as our planned pipeline - will inform you on this thread when it's ready

PR #1200, opened 2023-03-10
feat: `json` & `jsonb` and their array type support

v3g42:
Issue to track: https://github.com/getdozer/dozer/issues/956#issuecomment-1519154243, https://github.com/getdozer/dozer/issues/1228#issue-1620709773

Screenshot 2023-04-27 at 4 26 10 PM

gRPC
<details>

```json
{
    "records": [
        {
            "id": "0",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "1"
            }
        },
        {
            "id": "1",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "3",
                "jsonb": {
                    "number_value": 1
                },
                "json": {
                    "number_value": 1
                }
            }
        },
        {
            "id": "2",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "7",
                "jsonb": {
                    "struct_value": {
                        "fields": []
                    }
                },
                "json": {
                    "struct_value": {
                        "fields": []
                    }
                }
            }
        },
        {
            "id": "3",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "2",
                "jsonb": {
                    "list_value": {
                        "values": [
                            {
                                "string_value": "foo"
                            },
                            {
                                "string_value": "bar"
                            },
                            {
                                "string_value": "baz"
                            }
                        ]
                    }
                },
                "json": {
                    "list_value": {
                        "values": [
                            {
                                "string_value": "foo"
                            },
                            {
                                "string_value": "bar"
                            },
                            {
                                "string_value": "baz"
                            }
                        ]
                    }
                }
            }
        },
        {
            "id": "4",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "0",
                "jsonb": {
                    "list_value": {
                        "values": [
                            {
                                "number_value": 1
                            },
                            {
                                "number_value": 2
                            },
                            {
                                "number_value": 3
                            }
                        ]
                    }
                },
                "json": {
                    "list_value": {
                        "values": [
                            {
                                "number_value": 1
                            },
                            {
                                "number_value": 2
                            },
                            {
                                "number_value": 3
                            }
                        ]
                    }
                }
            }
        },
        {
            "id": "5",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "5",
                "jsonb": {
                    "struct_value": {
                        "fields": [
                            {
                                "key": "2",
                                "value": {
                                    "number_value": 2
                                }
                            },
                            {
                                "key": "3",
                                "value": {
                                    "number_value": 3
                                }
                            },
                            {
                                "key": "1",
                                "value": {
                                    "number_value": 1
                                }
                            }
                        ]
                    }
                },
                "json": {
                    "struct_value": {
                        "fields": [
                            {
                                "key": "2",
                                "value": {
                                    "number_value": 2
                                }
                            },
                            {
                                "key": "3",
                                "value": {
                                    "number_value": 3
                                }
                            },
                            {
                                "key": "1",
                                "value": {
                                    "number_value": 1
                                }
                            }
                        ]
                    }
                }
            }
        },
        {
            "id": "6",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "6",
                "jsonb": {
                    "struct_value": {
                        "fields": [
                            {
                                "key": "bar",
                                "value": {
                                    "struct_value": {
                                        "fields": [
                                            {
                                                "key": "2",
                                                "value": {
                                                    "number_value": 2
                                                }
                                            },
                                            {
                                                "key": "1",
                                                "value": {
                                                    "number_value": 1
                                                }
                                            },
                                            {
                                                "key": "3",
                                                "value": {
                                                    "number_value": 3
                                                }
                                            }
                                        ]
                                    }
                                }
                            },
                            {
                                "key": "baz",
                                "value": {
                                    "null_value": "NULL_VALUE"
                                }
                            },
                            {
                                "key": "foo",
                                "value": {
                                    "list_value": {
                                        "values": [
                                            {
                                                "number_value": 1
                                            },
                                            {
                                                "number_value": 2
                                            },
                                            {
                                                "number_value": 3
                                            }
                                        ]
                                    }
                                }
                            }
                        ]
                    }
                },
                "json": {
                    "struct_value": {
                        "fields": [
                            {
                                "key": "foo",
                                "value": {
                                    "list_value": {
                                        "values": [
                                            {
                                                "number_value": 1
                                            },
                                            {
                                                "number_value": 2
                                            },
                                            {
                                                "number_value": 3
                                            }
                                        ]
                                    }
                                }
                            },
                            {
                                "key": "bar",
                                "value": {
                                    "struct_value": {
                                        "fields": [
                                            {
                                                "key": "3",
                                                "value": {
                                                    "number_value": 3
                                                }
                                            },
                                            {
                                                "key": "2",
                                                "value": {
                                                    "number_value": 2
                                                }
                                            },
                                            {
                                                "key": "1",
                                                "value": {
                                                    "number_value": 1
                                                }
                                            }
                                        ]
                                    }
                                }
                            },
                            {
                                "key": "baz",
                                "value": {
                                    "null_value": "NULL_VALUE"
                                }
                            }
                        ]
                    }
                }
            }
        },
        {
            "id": "7",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 0,
                "id": "4",
                "jsonb": {
                    "string_value": "foo"
                },
                "json": {
                    "string_value": "foo"
                }
            }
        },
        {
            "id": "8",
            "record": {
                "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
                "__dozer_record_version": 80,
                "id": "8",
                "jsonb": {
                    "list_value": {
                        "values": [
                            {
                                "struct_value": {
                                    "fields": [
                                        {
                                            "key": "digit",
                                            "value": {
                                                "number_value": 30
                                            }
                                        },
                                        {
                                            "key": "letter",
                                            "value": {
                                                "string_value": "A"
                                            }
                                        }
                                    ]
                                }
                            },
                            {
                                "struct_value": {
                                    "fields": [
                                        {
                                            "key": "letter",
                                            "value": {
                                                "string_value": "B"
                                            }
                                        },
                                        {
                                            "key": "digit",
                                            "value": {
                                                "number_value": 31
                                            }
                                        }
                                    ]
                                }
                            }
                        ]
                    }
                },
                "json": {
                    "list_value": {
                        "values": [
                            {
                                "struct_value": {
                                    "fields": [
                                        {
                                            "key": "digit",
                                            "value": {
                                                "number_value": 30
                                            }
                                        },
                                        {
                                            "key": "letter",
                                            "value": {
                                                "string_value": "A"
                                            }
                                        }
                                    ]
                                }
                            },
                            {
                                "struct_value": {
                                    "fields": [
                                        {
                                            "key": "letter",
                                            "value": {
                                                "string_value": "B"
                                            }
                                        },
                                        {
                                            "key": "digit",
                                            "value": {
                                                "number_value": 31
                                            }
                                        }
                                    ]
                                }
                            }
                        ]
                    }
                }
            }
        }
    ]
}
```
</details>

REST
<details>

```json
[
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 1,
        "jsonb": null,
        "json": null,
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 0,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 3,
        "jsonb": 1.0,
        "json": 1.0,
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 1,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 7,
        "jsonb": {},
        "json": {},
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 2,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 2,
        "jsonb": [
            "foo",
            "bar",
            "baz"
        ],
        "json": [
            "foo",
            "bar",
            "baz"
        ],
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 3,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 0,
        "jsonb": [
            1.0,
            2.0,
            3.0
        ],
        "json": [
            1.0,
            2.0,
            3.0
        ],
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 4,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 5,
        "jsonb": {
            "1": 1.0,
            "2": 2.0,
            "3": 3.0
        },
        "json": {
            "1": 1.0,
            "2": 2.0,
            "3": 3.0
        },
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 5,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 6,
        "jsonb": {
            "bar": {
                "1": 1.0,
                "2": 2.0,
                "3": 3.0
            },
            "baz": null,
            "foo": [
                1.0,
                2.0,
                3.0
            ]
        },
        "json": {
            "bar": {
                "1": 1.0,
                "2": 2.0,
                "3": 3.0
            },
            "baz": null,
            "foo": [
                1.0,
                2.0,
                3.0
            ]
        },
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 6,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 4,
        "jsonb": "foo",
        "json": "foo",
        "jsonb_array": null,
        "json_array": null,
        "__dozer_record_id": 7,
        "__dozer_record_version": 1
    },
    {
        "uuid": "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11",
        "id": 8,
        "jsonb": [
            {
                "digit": 30.0,
                "letter": "A"
            },
            {
                "digit": 31.0,
                "letter": "B"
            }
        ],
        "json": [
            {
                "digit": 30.0,
                "letter": "A"
            },
            {
                "digit": 31.0,
                "letter": "B"
            }
        ],
        "jsonb_array": [
            {
                "digit": 30.0,
                "letter": "A"
            },
            {
                "digit": 31.0,
                "letter": "B"
            }
        ],
        "json_array": [
            {
                "digit": 30.0,
                "letter": "A"
            },
            {
                "digit": 31.0,
                "letter": "B"
            }
        ],
        "__dozer_record_id": 8,
        "__dozer_record_version": 1
    }
]
```
</details>

For storing Json, we created our own json value type as following:
```rust
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord, Hash)]
enum JsonValue {
    Null,
    Bool(bool),
    Number(OrderedFloat<f64>),
    String(String),
    Array(Vec<JsonValue>),
    Object(BTreeMap<String, JsonValue>),
}
```
This differs from `serde_json::Value` only in the `Number` representation. We should be able to replace most `serde_json::Value` usage in our code with `JsonValue`.

For representation in API:
google already defined this for protobuf - https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L62


dozer-types/src/types/field.rs
```diff
@@ -14,7 +14,9 @@ pub const DATE_FORMAT: &str = "%Y-%m-%d";
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord, Hash)]
 pub enum Field {
     UInt(u64),
+    UInt128(u128),
```
snork-alt, line 17: let's wait to introduce these. There is going to be a lot of work to extend all math operators. Let's do it as a second step

dozer-types/src/types/field.rs
```diff
@@ -14,7 +14,9 @@ pub const DATE_FORMAT: &str = "%Y-%m-%d";
...
     Int(i64),
+    Int128(i128),
```
snork-alt, line 19: same as above

dozer-types/src/types/field.rs
```diff
@@ -23,7 +25,7 @@ pub enum Field {
     Decimal(Decimal),
     Timestamp(DateTime<FixedOffset>),
     Date(NaiveDate),
-    Bson(Vec<u8>),
+    Bson(serde_json::Value),
```
snork-alt, line 28: Let's rename it to `Json`

dozer-types/src/types/field.rs
```diff
@@ -476,18 +478,31 @@ impl<'a> FieldBorrow<'a> {
...
     String,
     Text,
     Binary,
     Decimal,
-    Timestamp,
-    Date,
+    Timestamp { unit: Option<TimeUnit> },
+    Date { offset: Option<FixedOffset> },
+    Duration { unit: Option<TimeUnit> },
     Bson,
```
snork-alt, line 506: same as above. rename to Json

dozer-types/src/types/field.rs
```diff
@@ -476,18 +478,31 @@ impl<'a> FieldBorrow<'a> {
...
     Boolean,
     String,
     Text,
     Binary,
     Decimal,
-    Timestamp,
-    Date,
+    Timestamp { unit: Option<TimeUnit> },
+    Date { offset: Option<FixedOffset> },
+    Duration { unit: Option<TimeUnit> },
```
chubei, line 505: I don't get it. `Field::Timestamp` and `Field::Duration` are number + unit. How is the unit optional?

dozer-types/src/types/field.rs
chloeminkyung: + agree

dozer-types/src/types/field.rs
chloeminkyung: set the default unit?

chubei: > set the default unit?

Time unit is always nanoseconds in Dozer now right? Are we going to support `Timestamp` and `Duration` in (for example) microseconds to save the 8 bytes?

v3g42:
As discussed, we will need to support all these types and we agreed to store the precision of date and time fields. 
Each of these changes will be done sequentially in different PRs. 

- [ ] Introduce Json
- [ ] Add TimeUnit to FieldType `DateTime`
- [ ] Add Offset to FieldType `Date`
- [ ] Introduce FieldType `Duration`
- [ ] Introduce `Time`

chloeminkyung:
few problems to be solved:

```rust
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord, Hash)]
pub enum Field {
    ...
    Json(Value),
    ....
```

- [ ] can't compare `serde_json::Value` with `_` [E0277] Help: the trait `std::cmp::PartialOrd<_>` is not implemented for `serde_json::Value`
- [ ] the trait bound `serde_json::Value: std::cmp::Ord` is not satisfied
- [ ] the trait bound `serde_json::Value: std::hash::Hash` is not satisfied

```rust
pub fn map_record_to_arrow(
    rec: Record,
    schema: &Schema,
) -> Result<RecordBatch, arrow::error::ArrowError> {
    let mut rows = vec![];

    for (idx, f) in rec.values.iter().enumerate() {
        let fd = schema.fields.get(idx).unwrap();
        let r = match (f, fd.typ) {
            ...
            (Field::Json(v), FieldType::Json) => {
                let keys = vec!["a", "b", "c", "d", "e", "f", "g", "h"];
                let values_data = UInt32Array::from(vec![0u32, 10, 20, 30, 40, 50, 60, 70]);
                // Construct a buffer for value offsets, for the nested array:
                //  [[a, b, c], [d, e, f], [g, h]]
                let entry_offsets = [0, 3, 6, 8];

                let map_array = MapArray::new_from_strings(
                    keys.clone().into_iter(),
                    &values_data,
                    &entry_offsets,
                )
                .unwrap();
                Arc::new(arrow_array::MapArray::from(ArrayData::from(map_array))) as ArrayRef
                // Arc::new(as_map_array(serde_json::Map::from(v.as_object().unwrap()))) as ArrayRef
                // Arc::new(arrow_types::Field::from(Field::Json(v.to_owned()))) as ArrayRef
                // Arc::new(DataType::Map(Box::from(Field::Json(v.to_owned())), false)) as ArrayRef
            }
           ....
```
- [ ] `map_record_to_arrow()` to fulfill `Field::Json`


```rust
pub fn map_field_type(typ: FieldType, metadata: Option<&mut HashMap<String, String>>) -> DataType {
    match typ {
        ....
        FieldType::Json => {
            metadata.map(|m| m.insert("logical_type".to_string(), "Json".to_string()));
            let field = ArrowField::new("null", DataType::Null, false);
            DataType::Map(Box::from(field), false)
        }
       ....
```
- [ ] `map_field_type()` to fulfill `Field::Json`

chubei:
Thinking more about it, I'd like to keep using `Field::Json(Vec<u8>)` instead of `Field::Json(serde_json::Value)`.

Pros:
- `Vec<u8>` can be (de)serialized and compared (`Ord`) trivially, saving time when building and querying from the cache.

Cons:
- In the future when we implement sql operators on `Json`, processors will have to parse the `Vec<u8>`.

chubei:
Another question, how do we expose `Json` in APIs. Here's what I think:

- Common grpc: `bytes`
- Typed grpc: `bytes`
- REST: Parsed `serde_json::Value`

chloeminkyung:
> Another question, how do we expose `Json` in APIs. Here's what I think:
> 
> * Common grpc: `bytes`
> * Typed grpc: `bytes`
> * REST: Parsed `serde_json::Value`
> * Python: Parsed `dict`/`list`/`string`/...

common grpc bytes? wouldn't it be better if we expose as string? readability wise

chubei:
> expose as string? readability wise

Not sure if the benefit is worth the parsing and serialization cost. I'm OK with either.

dozer-api/src/generator/oapi/utils.rs
```diff
@@ -140,7 +140,14 @@ fn convert_cache_type_to_schema_type(field_type: dozer_types::types::FieldType)
                 ..Default::default()
             })
         }
-        FieldType::Binary | FieldType::Bson => Type::Array(ArrayType {
+        FieldType::Json => Type::Object(ObjectType {
```
chubei, line 143: Json is not necessarily an object. Is there some kind of `Any`?

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+pub fn json_value_to_prost_kind(val: JsonValue) -> ProstValue {
+    ProstValue {
+        kind: match val {
+            JsonValue::Null => Some(Kind::NullValue(0)),
+            JsonValue::Bool(b) => Some(Kind::BoolValue(b)),
+            JsonValue::Number(n) => Some(Kind::NumberValue(*n)),
+            JsonValue::String(s) => Some(Kind::StringValue(s)),
+            JsonValue::Array(a) => {
+                let values: prost::alloc::vec::Vec<ProstValue> = a
+                    .iter()
```
chubei, line 120: Use `into_iter` and avoid cloning

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+                    .map(|val| json_value_to_prost_kind(val.to_owned()))
+                    .collect();
+                Some(Kind::ListValue(ListValue { values }))
+            }
+            JsonValue::Object(o) => {
+                let fields: prost::alloc::collections::BTreeMap<
+                    prost::alloc::string::String,
+                    ProstValue,
+                > = o
+                    .iter()
```
chubei, line 130: Use `into_iter` and avoid cloning

dozer-ingestion/src/connectors/grpc/adapter/default.rs
```diff
@@ -142,8 +142,10 @@ fn map_record(rec: grpc_types::types::Record, schema: &Schema) -> Result<Record,
             ) => Ok(dozer_types::types::Field::Text(a.clone())),
             (
                 grpc_types::types::value::Value::BytesValue(a),
-                dozer_types::types::FieldType::Bson,
-            ) => Ok(dozer_types::types::Field::Bson(a.clone())),
+                dozer_types::types::FieldType::Json,
```
chubei, line 145: I think we should expect an `google.protobuf.Value` here and convert `google.protobuf.Value` to `JsonValue`, just like we did in typed grpc service, but the other way around

dozer-ingestion/src/connectors/kafka/debezium/mapper.rs
```diff
@@ -114,9 +116,9 @@ fn convert_value(
                     })
                 }
                 "io.debezium.time.MicroTime" => Ok(Field::Null),
-                "io.debezium.data.Json" => value
-                    .as_str()
-                    .map_or(Ok(Field::Null), |s| Ok(Field::Bson(s.as_bytes().to_vec()))),
+                "io.debezium.data.Json" => value.as_str().map_or(Ok(Field::Null), |s| {
```
chubei, line 119: I'm not sure about this convesion. @karolisg can you have a look?

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -74,7 +74,11 @@ pub fn postgres_type_to_field(
                     .unwrap();
                     Ok(Field::from(date))
                 }
-                Type::JSONB | Type::JSON => Ok(Field::Bson(v.to_vec())),
+                Type::JSONB | Type::JSON => {
```
chubei, line 77: Do we support `Type::JSON`? In the type conversion I only see `Type::JSONB`.

dozer-ingestion/src/connectors/kafka/debezium/mapper.rs
```diff
@@ -265,12 +269,14 @@ mod tests {
             Field::from(current_date),
             None
         );
-        let json_bytes = "{\"abc\":123}".as_bytes().to_vec();
+        let _json_bytes = "{\"abc\":123}".as_bytes().to_vec();
```
chubei, line 272: Remove this variable if unused?

dozer-types/src/types/field.rs
```diff
@@ -144,7 +145,7 @@ impl Field {
                 std::str::from_utf8(val)?,
                 DATE_FORMAT,
             )?)),
-            12 => Ok(Field::Bson(val.to_vec())),
+            12 => Ok(Field::Json(bincode::deserialize(val).unwrap())),
```
chubei, line 148: Return `Err` instead of `unwrap`

dozer-types/src/types/field.rs
```diff
@@ -261,9 +262,9 @@ impl Field {
         }
     }
 
-    pub fn as_bson(&self) -> Option<&[u8]> {
+    pub fn as_json(&self) -> Option<JsonValue> {
```
chubei, line 265: I think `as_*` methods are not supposed to do cloning, let's return `Option<&JsonValue>`

dozer-types/src/types/field.rs
```diff
@@ -47,7 +48,7 @@ impl Field {
             Field::Decimal(_) => 16,
             Field::Timestamp(_) => 8,
             Field::Date(_) => 10,
-            Field::Bson(b) => b.len(),
+            Field::Json(b) => bincode::serialize(b).unwrap().len(),
```
chubei, line 51: Leaving a comment to remind myself to think of something better

dozer-types/src/types/field.rs
```diff
@@ -600,8 +601,8 @@ pub enum FieldType {
     Timestamp,
     /// Allows for every date from Jan 1, 262145 BCE to Dec 31, 262143 CE.
     Date,
-    /// Bytes that are valid BSON.
-    Bson,
+    /// Bytes that are valid JSON.
```
chubei, line 604: We're not storing bytes, the the comment can just be `JsonValue`

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+                        )
+                    })
+                    .collect();
+                Some(Kind::StructValue(Struct { fields }))
+            }
+        },
+    }
+}
+
+// todo: not sure whether we need to involve serde_json conversion, and From<serde_json> for try_get from row
```
chubei, line 144: Regarding the `todo` comment, I think we do need this conversion

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+pub fn serde_json_to_json_value(value: Value) -> JsonValue {
+    match value {
+        Value::Null => JsonValue::Null,
+        Value::Bool(b) => JsonValue::Bool(b),
+        Value::Number(n) => JsonValue::Number(OrderedFloat(n.as_f64().unwrap())),
```
chubei, line 149: Don't `unwrap` here. Return an appropriate `Err` saying that we don't support json numbers out of the `f64` range.

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+        Value::String(s) => JsonValue::String(s),
+        Value::Array(a) => JsonValue::Array(
+            a.iter()
```
chubei, line 152: Use `into_iter` and avoid cloning

dozer-types/src/json_types.rs
```diff
@@ -1,8 +1,55 @@
...
+        }
+    }
+}
+
+// todo: for serde_json conversion
+impl FromStr for JsonValue {
+    type Err = SerializationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let object: Value = serde_json::from_str(s).unwrap();
```
chubei, line 49: Don't `unwrap` and return `Err`. The `FromStr::Err` type should be `DeserializationError`.

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
         Field::Date(n) => Value::String(n.format(DATE_FORMAT).to_string()),
-        Field::Bson(b) => Value::from(b),
+        Field::Json(b) => json_value_to_serde_json(b),
         Field::Point(point) => convert_x_y_to_object(&point.0.x_y()),
         Field::Duration(d) => convert_duration_to_object(&d),
         Field::Null => Value::Null,
     }
 }
 
+pub fn json_value_to_serde_json(value: JsonValue) -> Value {
```
chubei, line 90: We need to get rid of `json_value_to_serde_json` and `serde_json_to_json_value` methods.

All use of `serde_json::Value` can be removed if we properly implement `Serialize` and `Deserialize` for `JsonValue`, but let's leave it in the next PR.

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,89 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+    match value {
+        JsonValue::Null => Value::Null,
+        JsonValue::Bool(b) => Value::Bool(b),
+        JsonValue::Number(n) => Value::Number(Number::from_f64(*n).unwrap()),
+        JsonValue::String(s) => Value::String(s),
+        JsonValue::Array(a) => Value::Array(
+            a.iter()
```
chubei, line 97: Use `into_iter` and avoid cloning

dozer-tests/src/e2e_tests/checker/client.rs
```diff
@@ -361,7 +361,7 @@ fn oapi_type_matches(oapi_type: &dozer_api::openapiv3::Type, field_type: FieldTy
                 true
             }
         }
-        (Array(array_type), FieldType::Binary | FieldType::Bson) => {
+        (Array(array_type), FieldType::Binary | FieldType::Json) => {
```
chubei, line 364: This check needs to be consistent with oapi documentation generation

dozer-log-python/src/mapper.rs
```diff
@@ -76,7 +76,7 @@ fn map_value(value: Field, py: Python) -> PyResult<Py<PyAny>> {
         Field::Decimal(v) => Ok(v.to_string().to_object(py)),
         Field::Timestamp(v) => Ok(v.to_string().to_object(py)),
         Field::Date(v) => Ok(v.to_string().to_object(py)),
-        Field::Bson(v) => Ok(v.to_object(py)),
+        Field::Json(v) => Ok(v.to_string().to_object(py)),
```
chubei, line 79: Map `Json` properly:

- Null -> None
- Bool -> bool
- Number -> number
- Array -> list
- Object -> map

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -299,8 +296,11 @@ mod tests {
         );
 
         // UTF-8 bytes representation of json (https://www.charset.org/utf-8)
```
chubei, line 298: Remove outdated comment

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -74,7 +74,11 @@ pub fn postgres_type_to_field(
                     .unwrap();
                     Ok(Field::from(date))
                 }
-                Type::JSONB | Type::JSON => Ok(Field::Bson(v.to_vec())),
+                Type::JSONB | Type::JSON => {
+                    let json: JsonValue =
+                        serde_json_to_json_value(serde_json::from_slice(v).unwrap());
```
chubei, line 79: Don't `unwrap`.

chubei (changes requested):

dozer-types/src/json_types.rs
chloeminkyung: yup

dozer-types/src/json_types.rs
chloeminkyung: yup

dozer-ingestion/src/connectors/grpc/adapter/default.rs
chloeminkyung: got it!

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: we do have support for `Type::JSON`, `Type::JSONB`, `Type::JSON_ARRAY`, `Type::JSONB_ARRAY`

dozer-ingestion/src/connectors/kafka/debezium/mapper.rs
chloeminkyung: yup

dozer-ingestion/src/connectors/postgres/helper.rs
chubei: In `postgres_type_to_dozer_type` I only see `Type::JSONB`

dozer-types/src/types/field.rs
chloeminkyung: yup!

dozer-types/src/types/field.rs
chloeminkyung: yup!

dozer-types/src/types/field.rs
chloeminkyung: ok!

dozer-types/src/types/field.rs
chloeminkyung: okok

dozer-types/src/json_types.rs
chloeminkyung: ok

dozer-types/src/json_types.rs
chloeminkyung: got it

dozer-types/src/json_types.rs
chloeminkyung: Ok!

dozer-types/src/json_types.rs
chloeminkyung: ok!

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: ok!

dozer-tests/src/e2e_tests/checker/client.rs
chloeminkyung: ok!

dozer-log-python/src/mapper.rs
chloeminkyung: ok!

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: yup

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: here it is :) - https://github.com/getdozer/dozer/pull/1200/files#diff-23b2d6cb5efb3cee96f64fc2eb5b987363d9099ea56686f03935d41d1ebddb1aR109

dozer-types/src/json_types.rs
```diff
@@ -33,13 +79,105 @@ pub fn field_to_json_value(field: Field) -> Value {
...
         Field::Duration(d) => convert_duration_to_object(&d),
         Field::Null => Value::Null,
     }
 }
 
+pub fn json_value_to_serde_json(value: JsonValue) -> Value {
+    match value {
+        JsonValue::Null => Value::Null,
+        JsonValue::Bool(b) => Value::Bool(b),
+        JsonValue::Number(n) => Value::Number(Number::from_f64(*n).unwrap()),
```
chubei, line 93: We should handle `NaN` and `Inf` instead of `unwrap`

chubei (changes requested):
Great! Only a few more to go!

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -74,7 +74,15 @@ pub fn postgres_type_to_field(
                     .unwrap();
                     Ok(Field::from(date))
                 }
-                Type::JSONB | Type::JSON => Ok(Field::Bson(v.to_vec())),
+                Type::JSONB | Type::JSON | Type::JSONB_ARRAY | Type::JSON_ARRAY => {
+                    let val = serde_json::from_slice(v).map_err(|_| {
+                        PostgresSchemaError::ColumnTypeNotSupported(column_type.name().to_string())
```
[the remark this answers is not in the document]
chloeminkyung: got it

dozer-ingestion/tests/test_suite/connectors/sql.rs
```diff
@@ -72,8 +72,8 @@ pub fn create_table_with_all_supported_data_types(table_name: &str) -> String {
             '1970-01-01 00:00:00',
             0,
             0,
-            '{{}}'::json,
-            '{{}}'::json,
+            '{{}}'::jsonb,
```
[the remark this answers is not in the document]
chloeminkyung: added :)

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -147,14 +155,10 @@ pub fn value_to_field(
...
-        &Type::JSONB => {
-            let value: Result<serde_json::Value, _> = row.try_get(idx);
-
-            value.map_or_else(handle_error, |v| {
-                Ok(Field::Bson(
-                    bson::to_vec(&v).map_err(|e| JSONBParseError(e.to_string()))?,
-                ))
-            })
+        &Type::JSONB | &Type::JSON => convert_row_value_to_field!(row, idx, serde_json::Value),
+        &Type::JSONB_ARRAY | &Type::JSON_ARRAY => {
```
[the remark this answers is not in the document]
chloeminkyung: because `*_ARRAY` type cannot be natively converted to `serde_json::Value` as the macro that we are using is like followed:

```rust
macro_rules! convert_row_value_to_field {
    ($a:ident, $b:ident, $c:ty) => {{
        let value: Result<$c, _> = $a.try_get($b);
        value.map_or_else(handle_error, |val| Ok(Field::from(val)))
    }};
}
```
what we need for `*_ARRAY` is `Vec<serde_json::Value>` so had to parse differently. However, in `postgres_type_to_field` we don't specify `serde_json::Value` while deserialize from bytes using `from_slice`. In this case, it automatically gets its type as `Vec<serde_json::Value>` in case it is `*_ARRAY`.

dozer-log-python/src/mapper.rs
```diff
@@ -76,13 +80,32 @@ fn map_value(value: Field, py: Python) -> PyResult<Py<PyAny>> {
...
+        JsonValue::Number(n) => n.to_object(py),
+        JsonValue::String(s) => s.to_object(py),
+        JsonValue::Array(a) => {
+            PyList::new(py, a.into_iter().map(|val| map_json_py(val, py))).to_object(py)
+        }
+        JsonValue::Object(o) => {
+            let obj = PyDict::new(py);
+            let _ = o
+                .into_iter()
+                .map(|(key, val)| obj.set_item(key, map_json_py(val, py)));
```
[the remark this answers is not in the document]
chloeminkyung: then we have to use unwrap() is it ok?

dozer-types/src/field_type.rs
```diff
@@ -131,3 +132,20 @@ impl From<DozerDuration> for Field {
...
+        Field::Json(serde_json_to_json_value(value).unwrap_or(JsonValue::Null))
+    }
+}
+
+impl From<Vec<serde_json::Value>> for Field {
+    fn from(value: Vec<serde_json::Value>) -> Self {
+        Field::Json(JsonValue::Array(
+            value
+                .into_iter()
+                .map(|val| serde_json_to_json_value(val).unwrap_or(JsonValue::Null))
```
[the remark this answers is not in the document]
chloeminkyung: then we need to change the `impl From<serde_json::Value> for Field` 's return type as `Result<Field, E>` but then it doesn't inherit the original `From` implementation that we want to follow, or should we panic?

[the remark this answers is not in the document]
chubei: We don't need to implement `From`. We already have `serde_json_value_to_field_value`.

dozer-types/src/json_types.rs
chloeminkyung: can't we just do `JsonValue::Number(n) => json!(*n),` then ?

dozer-log-python/src/mapper.rs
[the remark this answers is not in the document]
chubei: I don't get it. Why `for_each` leads to `unwrap`?

dozer-types/src/field_type.rs
```diff
@@ -131,3 +132,20 @@ impl From<DozerDuration> for Field {
         Field::Duration(v)
     }
 }
+
+impl From<serde_json::Value> for Field {
+    fn from(value: serde_json::Value) -> Self {
+        Field::Json(serde_json_to_json_value(value).unwrap_or(JsonValue::Null))
```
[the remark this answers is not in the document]
chubei: We don't need to implement From. We already have serde_json_value_to_json_value.

dozer-types/src/json_types.rs
chubei: I guess it also panics. You can try it.

dozer-ingestion/src/connectors/kafka/debezium/mapper.rs
karolisg: Looks correct. Maybe we can have proper error `JsonValue::from_str(s).unwrap()` instead of unwrap?

dozer-api/src/generator/oapi/utils.rs
chloeminkyung: yes will refactor the function to return the `SchemaKind` instead of `Type` itself as `Type` doesn't have anything that can be able to be mapped into `Any`

dozer-tests/src/e2e_tests/checker/client.rs
chloeminkyung: matched to check on the `SchemaKind`

dozer-ingestion/src/connectors/postgres/helper.rs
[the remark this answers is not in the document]
chloeminkyung: Ah got your point, then `postgres_type_to_field` need to be changed, because in case of change this here itself as followed, 

```rust
 &Type::JSONB | &Type::JSON | &Type::JSONB_ARRAY | &Type::JSON_ARRAY => {
            convert_row_value_to_field!(row, idx, serde_json::Value)
        }
```
then will get this error 

```
ERROR Value conversion error: cannot convert between the Rust type `serde_json::value::Value` and the Postgres type `_jsonb`
```

dozer-types/src/field_type.rs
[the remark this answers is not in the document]
chloeminkyung: ah got it!

dozer-types/src/field_type.rs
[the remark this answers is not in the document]
chloeminkyung: ah got it!

dozer-types/src/json_types.rs
chloeminkyung:
```sql
insert into test_uuid values (
    'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11',
    9,
    'Infinity'::jsonb,
    'Nan'::json,
    null,
    null
);
```

for `Infinity` -> Illegal floating point literal 
for `Nan` -> JSON standard does not allow such tokens

dozer-ingestion/src/connectors/kafka/debezium/mapper.rs
chloeminkyung: ok! thanks @karolisg @chubei

dozer-ingestion/src/errors.rs
```diff
@@ -374,6 +374,9 @@ pub enum DebeziumSchemaError {
     #[error("Invalid date")]
     InvalidDateError,
 
+    #[error("Invalid json")]
+    InvalidJsonError,
```
chubei, line 378:
```suggestion
    #[error("Invalid json: {0}")]
    InvalidJsonError(#[source] serde_json::Error),
```

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -74,7 +74,40 @@ pub fn postgres_type_to_field(
...
-                Type::JSONB | Type::JSON => Ok(Field::Bson(v.to_vec())),
+                Type::JSONB | Type::JSON => {
+                    let val: serde_json::Value = serde_json::from_slice(v).map_err(|_| {
+                        PostgresSchemaError::JSONBParseError(format!(
+                            "Error converting to a single row for: {}",
+                            column_type.name()
+                        ))
+                    })?;
+                    let json: JsonValue = serde_json_to_json_value(val).map_err(|_| {
+                        PostgresSchemaError::ColumnTypeNotSupported(column_type.name().to_string())
```
chubei, line 85: This is not `ColumnTypeNotSupported`, maybe we can forward the `TypeError`?

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -74,7 +74,40 @@ pub fn postgres_type_to_field(
...
+                            "Error converting to a single row for: {}",
+                            column_type.name()
+                        ))
+                    })?;
+                    let json: JsonValue = JsonValue::Array(
+                        val.into_iter()
+                            .map(|v| {
+                                serde_json_to_json_value(v)
+                                    .map_err(|_| {
+                                        PostgresSchemaError::ColumnTypeNotSupported(
```
chubei, line 101: This is not ColumnTypeNotSupported, maybe we can forward the TypeError?

dozer-ingestion/src/connectors/postgres/helper.rs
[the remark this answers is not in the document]
chubei: I still don't get it. If some bytes can be successfully deserialized to `Vec<serde_json::Value>` in JSON format, then by definition it's also a `serde_json::Value`, because an array of JSON value is a JSON value. Can you try to find the root cause of the failure?

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -147,15 +180,30 @@ pub fn value_to_field(
...
             })
         }
+        &Type::JSONB_ARRAY | &Type::JSON_ARRAY => {
+            let value: Result<Vec<serde_json::Value>, _> = row.try_get(idx);
+            value.map_or_else(handle_error, |val| {
+                Ok(Field::Json(JsonValue::Array(
+                    val.into_iter()
+                        .map(|v| {
+                            serde_json_to_json_value(v)
+                                .map_err(|_| ColumnTypeNotSupported(col_type.name().to_string()))
```
chubei, line 200: This is not ColumnTypeNotSupported, maybe we can forward the TypeError?

dozer-ingestion/tests/test_suite/connectors/sql.rs
```diff
@@ -146,7 +164,8 @@ fn field_type_to_sql(field_type: FieldType) -> Option<String> {
         FieldType::Decimal => Some("NUMERIC".to_string()),
         FieldType::Timestamp => Some("TIMESTAMP".to_string()),
         FieldType::Date => Some("DATE".to_string()),
-        FieldType::Bson => Some("JSONB".to_string()),
+        // todo: mapping for `JSON`, `JSONB[]`, `JSON[]` is needed
```
chubei, line 167: This is OK. We only need to chooose one format in postgres that can represent `FieldType::Json`. This function is for CUD test, not data type test.

dozer-log-python/src/mapper.rs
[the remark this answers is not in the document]
chubei: `map` without being consumed does nothing, so this function will always return an empty python dict. This is a bug.

dozer-tests/src/e2e_tests/checker/client.rs
```diff
@@ -361,7 +373,8 @@ fn oapi_type_matches(oapi_type: &dozer_api::openapiv3::Type, field_type: FieldTy
                 true
             }
         }
-        (Array(array_type), FieldType::Binary | FieldType::Bson) => {
+        (Array(_), FieldType::Json) | (Object(_), FieldType::Json) => true,
```
chubei, line 376: Remove this line because `FiledType::Json` doesn't match any `Type`

dozer-types/src/json_types.rs
chubei: > for Infinity -> Illegal floating point literal
> for Nan -> JSON standard does not allow such tokens

This is exactly how JSON works. What do you mean by "do we need it"?

chubei: I guess `json!(*n)` panics if `n` is `NaN` of `Inf`. We need to handle this error instead of panic.

chloeminkyung: no what i mean is there is no way that we are getting NaN of Inf as it's already considered illegal from db side

dozer-ingestion/src/connectors/postgres/helper.rs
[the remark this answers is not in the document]
chloeminkyung: yes the cause is due to

```rust
let value: Result<$c, _> = $a.try_get($b);
```

in the macro, it cannot be getting try_get a row which is serde_json::Value which is already an Array type

dozer-log-python/src/mapper.rs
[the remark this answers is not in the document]
chloeminkyung: i don't get it - the op that we want to do is just `set_item`, what's wrong with that?

dozer-ingestion/tests/test_suite/connectors/sql.rs
chloeminkyung: ok

dozer-types/src/json_types.rs
chubei: The problem is `JsonValue::Number` is public, and it's very possible that it will be constructed in other places, in addition to in the connector. And that "other places" may produce `NaN` and put it in `JsonValue::Number`. This leaves us with three options:

1. Check if that `f64`  is finite and return error if not.
2. Don't check and panic.
3. Create a new type `FiniteF64` and use it instead of `f64` in `JsonValue::Number`.

Option 3 is best but also the most difficult. Option 2 is the easiest but the worst. Option 1 is just a way to avoid panicking but still doesn't go all the way into implementing a new type.

dozer-log-python/src/mapper.rs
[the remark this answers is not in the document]
chubei: I'm not sure if I misunderstood the code, and I apologize if you already read this before, but please have a look at https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map, where it says

> `map()` is lazy

The closure that gets passed to `map` doesn't get called unless the return value of `map` is consumed. By consuming I mean calling `Iterator::next`, which can be a `for` loop, a `collect`, or something else.

I don't see any consuming happening here, so I believe `obj.set_item` is never called.

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: changed

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: changed

dozer-tests/src/e2e_tests/checker/client.rs
chloeminkyung: agreed

dozer-ingestion/src/connectors/postgres/helper.rs
chloeminkyung: yup!

dozer-ingestion/src/connectors/postgres/helper.rs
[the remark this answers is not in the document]
chubei: Got it! Thanks!

dozer-types/src/json_types.rs
chloeminkyung: sounds good - will proceed with option 3

chubei (changes requested):

dozer-ingestion/src/errors.rs
```diff
@@ -374,12 +377,17 @@ pub enum DebeziumSchemaError {
...
+    #[error("Invalid json: {0}")]
+    InvalidJsonError(String),
+
     // #[error("Invalid time")]
     // InvalidTimeError,
     #[error("Invalid timestamp")]
     InvalidTimestampError,
 }
 
+impl Eq for DebeziumSchemaError {}
```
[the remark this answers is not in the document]
chloeminkyung: cannot be because of 
```
the trait bound `dozer_types::rust_decimal::Error: std::cmp::Eq` is not satisfied [E0277] Note: required by a bound in `std::cmp::AssertParamIsEq`
```

dozer-log-python/src/mapper.rs
```diff
@@ -76,13 +79,32 @@ fn map_value(value: Field, py: Python) -> PyResult<Py<PyAny>> {
...
+        JsonValue::Bool(b) => b.to_object(py),
+        JsonValue::Number(n) => n.to_object(py),
+        JsonValue::String(s) => s.to_object(py),
+        JsonValue::Array(a) => {
+            PyList::new(py, a.into_iter().map(|val| map_json_py(val, py))).to_object(py)
+        }
+        JsonValue::Object(o) => {
+            let obj = PyDict::new(py);
+            for (key, val) in o {
+                obj.set_item(key, map_json_py(val, py)).unwrap();
```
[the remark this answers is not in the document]
chloeminkyung: then should we `expect`?

dozer-ingestion/src/errors.rs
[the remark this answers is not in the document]
chubei: If we can't derive, we shouldn't `impl`. Why is it needed?

dozer-log-python/src/mapper.rs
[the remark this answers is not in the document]
chubei: No I actually meant don't panic. Return `Err`.

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -74,7 +75,40 @@ pub fn postgres_type_to_field(
                     .unwrap();
                     Ok(Field::from(date))
                 }
-                Type::JSONB | Type::JSON => Ok(Field::Bson(v.to_vec())),
+                Type::JSONB | Type::JSON => {
```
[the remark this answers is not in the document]
chloeminkyung: added!

dozer-ingestion/src/errors.rs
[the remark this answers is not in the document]
chloeminkyung: Cause this has been implemented - https://github.com/getdozer/dozer/pull/1200#discussion_r1179868789

dozer-ingestion/src/connectors/postgres/helper.rs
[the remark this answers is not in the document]
chloeminkyung: added!

dozer-log-python/src/mapper.rs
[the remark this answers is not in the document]
chloeminkyung: got it

dozer-ingestion/src/errors.rs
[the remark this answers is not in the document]
chloeminkyung: fixed!

dozer-log-python/src/mapper.rs
```diff
@@ -76,13 +80,37 @@ fn map_value(value: Field, py: Python) -> PyResult<Py<PyAny>> {
...
+        JsonValue::String(s) => Ok(s.to_object(py)),
+        JsonValue::Array(a) => Ok(PyList::new(
+            py,
+            a.into_iter().map(|val| map_json_py(val, py).unwrap()),
+        )
+        .to_object(py)),
+        JsonValue::Object(o) => {
+            let obj = PyDict::new(py);
+            for (key, val) in o {
+                obj.set_item(key, map_json_py(val, py).unwrap()).map_or(
```
chubei, line 104: Use `?` instead of `unwrap`.

Can we also use `?` instead of `map_or`?

chubei (changes requested):

dozer-log-python/src/mapper.rs
chloeminkyung: right right fixed!

dozer-log-python/src/mapper.rs
```diff
@@ -76,13 +79,34 @@ fn map_value(value: Field, py: Python) -> PyResult<Py<PyAny>> {
...
 
+fn map_json_py(val: JsonValue, py: Python) -> PyResult<Py<PyAny>> {
+    match val {
+        JsonValue::Null => Ok(py.None()),
+        JsonValue::Bool(b) => Ok(b.to_object(py)),
+        JsonValue::Number(n) => Ok(n.to_object(py)),
+        JsonValue::String(s) => Ok(s.to_object(py)),
+        JsonValue::Array(a) => Ok(PyList::new(
+            py,
+            a.into_iter().map(|val| map_json_py(val, py).unwrap()),
```
chubei, line 97: Caught one more `unwrap`!

dozer-types/src/arrow_types/from_arrow.rs
```diff
@@ -175,7 +175,7 @@ pub fn map_arrow_to_dozer_type(dt: &DataType) -> Result<FieldType, FromArrowErro
         // DataType::Dictionary(_, _) => {}
         // DataType::Decimal128(_, _) => {}
         // DataType::Decimal256(_, _) => {}
-        // DataType::Map(_, _) => {}
+        DataType::Map(_, _) => Ok(FieldType::Json),
```
chubei, line 178: If we don't implement conversion in `map_value_to_dozer_field`, we shouldn't do here.

dozer-types/src/arrow_types/to_arrow.rs
```diff
@@ -152,10 +152,7 @@ pub fn map_field_type(typ: FieldType, metadata: Option<&mut HashMap<String, Stri
             metadata.map(|m| m.insert("logical_type".to_string(), "Binary".to_string()));
             DataType::Binary
         }
-        FieldType::Bson => {
-            metadata.map(|m| m.insert("logical_type".to_string(), "Bson".to_string()));
-            DataType::Binary
-        }
+        FieldType::Json => todo!(),
```
chubei, line 155: Maybe we serialize JSON to arrow String and add a logical type? @v3g42 what do you think?

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,115 @@ pub fn field_to_json_value(field: Field) -> Value {
         Field::Decimal(n) => Value::String(n.to_string()),
         Field::Timestamp(ts) => Value::String(ts.to_rfc3339_opts(SecondsFormat::Millis, true)),
         Field::Date(n) => Value::String(n.format(DATE_FORMAT).to_string()),
-        Field::Bson(b) => Value::from(b),
+        Field::Json(b) => json_value_to_serde_json(b).unwrap(),
```
chubei, line 83: Let's propagate the error instead of `unwrap`.

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,115 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+            if n.0.is_finite() {
+                Ok(json!(n.0))
+            } else {
+                Err(DeserializationError::F64TypeConversionError)
+            }
+        }
+        JsonValue::String(s) => Ok(Value::String(s)),
+        JsonValue::Array(a) => Ok(Value::Array(
+            a.into_iter()
+                .map(|val| json_value_to_serde_json(val).unwrap())
```
chubei, line 104: Please don't `unwrap`. We make make use of the fact that `Result<Vec<T>, Error>` can be a `collect` output type and write

```rust
a.into_iter().map(|value| json_value_to_serde_json(val)).collect::<Result<Vec<_>, _>>()?
```

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,115 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+                .collect(),
+        )),
+        JsonValue::Object(o) => {
+            let mut values: Map<String, Value> = Map::new();
+            for (key, val) in o {
+                values.insert(key, json_value_to_serde_json(val).unwrap());
```
chubei, line 110: Please don't unwrap.

dozer-types/src/json_types.rs
```diff
@@ -33,13 +80,115 @@ pub fn field_to_json_value(field: Field) -> Value {
...
+        Value::Number(n) => Ok(JsonValue::Number(OrderedFloat(match n.as_f64() {
+            Some(f) => f,
+            None => return Err(DeserializationError::F64TypeConversionError),
+        }))),
+        Value::String(s) => Ok(JsonValue::String(s)),
+        Value::Array(a) => Ok(JsonValue::Array(
+            a.into_iter()
+                .map(|val| {
+                    serde_json_to_json_value(val)
+                        .map_err(|_| DeserializationError::F64TypeConversionError)
```
chubei, line 177: Please don't unwrap.

chubei (changes requested):

dozer-types/src/arrow_types/to_arrow.rs
v3g42: Yup makes sense.

chloeminkyung: so then to_arrow is JSON -> String but how are we handling the String -> Json in from arrow?, how to differentiate with the Normal String?

and @chubei last time you told me to return error for json what should we do?

dozer-log-python/src/mapper.rs
chloeminkyung: changed like how we handle JsonValue::Object type

dozer-types/src/json_types.rs
chloeminkyung: right will do that

dozer-types/src/json_types.rs
chloeminkyung: removed

dozer-types/src/json_types.rs
chloeminkyung: removed

dozer-types/src/json_types.rs
chubei: > an iterator over elements of type `serde_json::Value`

This is because of the `unwrap`.

Either way is fine.

dozer-types/src/arrow_types/to_arrow.rs
chubei: Let's go with converting to String. I believe it's better than returning an error.

When converting to arrow, we insert metadata saying the `logical_type` is `json`, and when converting from arrow, we check if that metadata is present and convert accordingly.

But let's do this in the next PR. In this PR we can just convert JSON to string without converting back.

chubei (approved):

PR #1213, opened 2023-03-11
fix: Restore docker publishing

mediuminvader:


.github/workflows/release.yaml
```diff
@@ -109,7 +109,11 @@ jobs:
             ${{ runner.os }}-cargo-release-
 
       - name: Build package
-        run: cargo build --release --target ${{ matrix.target }}
+        run: cargo build --features=snowflake --release --target ${{ matrix.target }}
```
chubei, line 112: Do we want snowflake?

.github/workflows/release.yaml
```diff
@@ -109,7 +109,11 @@ jobs:
...
+
+      - name: Build Linux binary
```
chubei, line 114: Why build twice?

dozer-tests/src/e2e_tests/checker/mod.rs
```diff
@@ -56,8 +56,10 @@ fn assert_command_fails(mut command: Command, expected_string_in_stdout: Option<
     if output.status.success() {
         panic!("Command {command:?} is expected to fail, but it succeeded");
     }
-    let stdout = String::from_utf8_lossy(&output.stdout);
-    error!("{}", stdout);
+    let stdout = String::from_utf8(output.stdout).unwrap();
```
chubei, line 59: This is my debug code. We should remove it.

chubei (changes requested):

.github/workflows/release.yaml
mediuminvader: I reverted the changes prior to my merge, we can remove if not needed

.github/workflows/release.yaml
mediuminvader: On the first run I got this error
`COPY failed: file not found in build context or excluded by .dockerignore: stat target/release/dozer: file does not exist`

dozer-tests/src/e2e_tests/checker/mod.rs
mediuminvader: removed

.github/workflows/release.yaml
mediuminvader: removed

.github/workflows/release.yaml
chubei: I think that's because the extra `--target x86_64-unknown-linux-gnu` parameter. We shouldn't build twice.

mediuminvader: I agree, what's the best solution?

chubei: Change `ci/Dockerfile` so it finds the build artifact?

mediuminvader: I removed --target since we're building on the specific architecture

chubei: I still don't get it. This line https://github.com/getdozer/dozer/pull/1213/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR123 has {{matrix.target}} in the path.

chubei (approved):

PR #1214, opened 2023-03-11
feat: introduce sqllogictest for dozer

xudong963:
Introduce sqllogictest for dozer, make it more flexible to add massive tests.

Ideally, the test set of sqlite can be reused later!

### DIFF
--- a/dozer-tests/Cargo.toml
+++ b/dozer-tests/Cargo.toml
@@ -14,6 +14,10 @@ path = "src/dozer_tests.rs"
 name = "dozer-test-client"
 path = "src/dozer_test_client.rs"
 
+[[bin]]
+name = "dozer-sqllogictest"
+path = "src/sqllogictest/src/main.rs"
+
 [dependencies]
 dozer-types = { path = "../dozer-types" }
 dozer-api = { path = "../dozer-api" }
@@ -25,6 +29,14 @@ futures = { version = "0.3.26", optional = true }
 env_logger = "0.10.0"
 clap = { version = "4.1.6", features = ["derive"] }
 dozer-orchestrator = { path = "../dozer-orchestrator" }
+rusqlite = { version = "0.28.0", features = ["bundled", "column_decltype"] }
+sqllogictest = "0.11.1"
+async-trait = "0.1.53"
+dozer-core = { path = "../dozer-core" }
+dozer-sql = { path = "../dozer-sql" }
+tempdir = "0.3.7"
+walkdir = "2.3.2"
+
 
 [dev-dependencies]
 sqlparser = "0.32.0"
--- a/dozer-tests/src/sqllogictest/suits/join.test
+++ b/dozer-tests/src/sqllogictest/suits/join.test
@@ -1,0 +1,308 @@
+control sortmode rowsort
+
+# actor
+statement ok
+CREATE TABLE actor(
+    actor_id integer NOT NULL,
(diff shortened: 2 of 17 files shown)


dozer-tests/Cargo.toml
```diff
@@ -25,6 +29,16 @@ futures = { version = "0.3.26", optional = true }
 env_logger = "0.10.0"
 clap = { version = "4.1.6", features = ["derive"] }
 dozer-orchestrator = { path = "../dozer-orchestrator" }
+rusqlite = { version = "0.28.0", features = ["bundled", "column_decltype"] }
+sqllogictest = "0.11.1"
+sqlparser = "0.32.0"
```
v3g42, line 34: Can you re-export sqlparser from `dozer-sql` ?

dozer-tests/Cargo.toml
```diff
@@ -25,6 +29,16 @@ futures = { version = "0.3.26", optional = true }
...
+thiserror = "1.0.38"
```
v3g42, line 35: This is already available through `dozer_types::thiserror`

dozer-tests/Cargo.toml
```diff
@@ -14,6 +14,10 @@ path = "src/dozer_tests.rs"
 name = "dozer-test-client"
 path = "src/dozer_test_client.rs"
 
+[[bin]]
+name = "dozer-sqllogictest"
+path = "src/sqllogictest/src/main.rs"
+
 [dependencies]
```
v3g42, line 21: Considering we are building a binary, can we restrict the dependencies with a certain feature flag ? So we can build it separately ?

v3g42 (changes requested):

v3g42:
Seems like a good idea. But it would be good to see some of the existing test cases ported before we adopt it widely.

snork-alt (changes requested):
Good start. But if we develop a new framework we need support for complex SQLs. On top of this, all previous test cases should be ported to the new framework.

For complex scenarios you can refer to this : https://github.com/getdozer/dozer-samples/blob/main/pg-flights/dozer-config.yaml

dozer-tests/src/sqllogictest/suits/join.test
```diff
@@ -0,0 +1,308 @@
...
+2 jack nicholson
+3 angelina jolie
+3 angelina jolie
+3 angelina jolie
+3 luigi guiness
+3 luigi guiness
+3 luigi guiness
+4 tom hanks
+
+# dozer has bug for the sql, less one row `1 penelope guiness`
```
xudong963, line 278: I find a left join bug  @mediuminvader

mediuminvader: Nice, but I'm refactoring the join right now #1180 , let's re run this on the next merge

dozer-tests/Cargo.toml
xudong963: It doesn't need many dependencies, so maybe we don't need to use a certain feature flag to restrict it

v3g42 (approved):
Thanks @xudong963. This looks good to me. 

I believe @mediuminvader wanted to incorporate a few more changes. i ll leave it you and him.

xudong963:
So let's merge？ Then let's iterate together? @mediuminvader 

I still have some ideas to solid new framework.

snork-alt (approved):

PR #1252, opened 2023-03-17
refactor: avg

chloeminkyung:
get rid of weighted avg

### DIFF
--- a/dozer-sql/src/pipeline/aggregation/avg.rs
+++ b/dozer-sql/src/pipeline/aggregation/avg.rs
@@ -1,11 +1,11 @@
-use crate::pipeline::aggregation::aggregator::{Aggregator};
+use crate::argv;
+use crate::pipeline::aggregation::aggregator::Aggregator;
 use crate::pipeline::aggregation::sum::{get_sum, SumState};
-use crate::pipeline::errors::PipelineError::{InvalidValue};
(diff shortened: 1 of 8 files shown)


dozer-sql/src/pipeline/aggregation/avg.rs
```diff
@@ -67,93 +78,79 @@ impl Aggregator for AvgAggregator {
...
-                let mut count =
-                    calculate_err_type!(Decimal::from_f64(0_f64), Avg, FieldType::Decimal);
-                for (field, cnt) in field_map {
-                    let cnt = calculate_err_field!(Decimal::from_u64(*cnt), Avg, field);
-                    sum += calculate_err_field!(field.to_decimal(), Avg, field) * cnt;
-                    count += cnt;
-                }
-                Ok(Field::Decimal(sum / count))
+            if *current_count == 0 {
+                return Ok(Field::UInt(0));
```
chubei, line 115: `Null` is better?

chloeminkyung: @snork-alt @mediuminvader feedback please!

chloeminkyung: yea I think it's right to change to `Field::Null`

PR #1282, opened 2023-03-21
test: Add sql integration tests

mediuminvader:


dozer-tests/src/e2e_tests/cases/sql_aggregations/expectations.json
```diff
@@ -0,0 +1,3 @@
+[
```
chubei, line 1: Do we want to add more expectations?

chubei (approved):

dozer-tests/src/e2e_tests/cases/sql_aggregations/expectations.json
mediuminvader: Yes, let me add em, I didn't notice the other cases

dozer-tests/src/e2e_tests/cases/sql-window-functions-sample/expectations.json
```diff
@@ -0,0 +1,40 @@
+[
+    "HealthyService",
+    {
```
chubei, line 3: Missing `hop` endpoint?

PR #1315, opened 2023-03-27
feat: u128 and i128 Field and FieldType

chloeminkyung:
- [x] implementation
- [x] unit tests


snork-alt (approved):

dozer-ingestion/tests/test_suite/connectors/arrow.rs
```diff
@@ -324,6 +326,17 @@ fn fields_to_arrow<'a, F: IntoIterator<Item = &'a Field>>(
             }
             Arc::new(builder.finish())
         }
+        FieldType::U128 => {
```
chubei, line 329: Just `panic` is OK.

dozer-types/src/helper.rs
```diff
@@ -75,6 +97,19 @@ impl Field {
                     })
                 }
             }
+            FieldType::U128 => {
```
chubei, line 100: I don't think we should convert `"null"` to `Field::Null`

chubei (commented):
Great work! Support looks fine inside dozer. I have some questions about conversions listed in the comments.

dozer-api/src/grpc/types_helper.rs
```diff
@@ -86,17 +85,22 @@ fn field_to_prost_value(f: Field) -> Value {
         Field::UInt(n) => Value {
             value: Some(value::Value::UintValue(n)),
         },
+        Field::U128(n) => Value {
```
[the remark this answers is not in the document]
chloeminkyung: TL;DR - No, it won't panic but we will lose data

`U128(n)`'s `n` is `u128` so if we do `as u64` there is extra data bits causing overflow, no, `as u64` handles ignoring extra bits

What should we do? cause there is no 1-to-1 mapping for `U128` in prost `Value`, should we create our own `U128` in that case?

dozer-ingestion/tests/test_suite/connectors/arrow.rs
```diff
@@ -268,7 +268,9 @@ pub fn record_batch_with_all_supported_data_types() -> arrow::record_batch::Reco
 fn field_type_to_arrow(field_type: FieldType) -> Option<arrow::datatypes::DataType> {
     match field_type {
         FieldType::UInt => Some(arrow::datatypes::DataType::UInt64),
+        FieldType::U128 => Some(arrow::datatypes::DataType::UInt64), // TODO: Map this correctly
```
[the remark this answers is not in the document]
chloeminkyung: Ok

dozer-api/src/grpc/types_helper.rs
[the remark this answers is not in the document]
chubei: Definitely. Either a `string` or a `message u128`. @mediuminvader @v3g42 which you think is better?

dozer-ingestion/tests/test_suite/connectors/arrow.rs
chloeminkyung: ok

dozer-api/src/grpc/types_helper.rs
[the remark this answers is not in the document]
v3g42: I think a message would be appropraite as there is metadata on the client side, to map it appropriately.

dozer-types/src/helper.rs
chloeminkyung: other types are all supporting in that way, I just followed the way we defined `UInt` and `Int`

dozer-api/src/grpc/types_helper.rs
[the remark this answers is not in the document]
chubei: `string` can also contain metadata, like
```proto
message Value {
  oneof Value {
    uint64 uint_value = 1;
    string uint_128_value = 2;
    ...
    string string_value = 6;
  }
}
```

dozer-types/src/helper.rs
chubei: Can you double check? Last time I checked it converting `"null"` to `Field::UInt` results in an `Err`.

dozer-api/src/grpc/types_helper.rs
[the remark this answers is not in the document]
v3g42: ok. i remember in code there is a reverse mapping of types back to Dozer Types purely based on data type. That needs to be addressed. I m ok with both.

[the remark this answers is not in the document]
chloeminkyung: so we go string?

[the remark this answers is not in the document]
chubei: I think `string` is better from human readbility standpoint. Let's do `string`.

dozer-types/src/helper.rs
chloeminkyung: https://github.com/getdozer/dozer/pull/1315/files/eafbddaef0509856831ef4022611164e3d4749b8#diff-6f3491eceadec0a62a4343f7e179d5b6eea02801eaa4efeb32e37880b2662f87R88

chubei: Great thanks!

dozer-api/src/generator/protoc/generator/implementation.rs
```diff
@@ -360,7 +360,9 @@ impl Names {
 fn convert_dozer_type_to_proto_type(field_type: FieldType) -> Result<String, GenerationError> {
     match field_type {
         FieldType::UInt => Ok("uint64".to_owned()),
+        FieldType::U128 => Ok("uint128".to_owned()),
```
chubei, line 363: This function must be consistent with `field_to_prost_value`.

dozer-api/src/rest/api_generator.rs
```diff
@@ -190,7 +190,9 @@ fn convert_x_y_to_object((x, y): &(OrderedFloat<f64>, OrderedFloat<f64>)) -> Val
 fn field_to_json_value(field: Field) -> Value {
     match field {
         Field::UInt(n) => Value::from(n),
+        Field::U128(n) => Value::from(n as u64), // TODO: Map this correctly
```
chubei, line 193: Let's also do `string` as in gRPC

chloeminkyung: yup

dozer-api/src/generator/protoc/generator/implementation.rs
chloeminkyung: yup

dozer-types/src/arrow_types/to_arrow.rs
```diff
@@ -127,13 +127,14 @@ pub fn map_record_to_arrow(
 pub fn map_field_type(typ: FieldType, metadata: Option<&mut HashMap<String, String>>) -> DataType {
     match typ {
         FieldType::UInt => DataType::UInt64,
+        FieldType::U128 => DataType::Utf8,
```
chubei, line 130: Arrow has `u128` and `i128`. We need to map to numeric types.

dozer-api/src/grpc/types_helper.rs
chubei, line 88: Should convert to string as discussed

chubei (changes requested):

chubei:
Please also update the table in `dozer-api/README.md`.

dozer-types/src/arrow_types/to_arrow.rs
chloeminkyung:
```
#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub enum DataType {
    /// Null type
    Null,
    /// A boolean datatype representing the values `true` and `false`.
    Boolean,
    /// A signed 8-bit integer.
    Int8,
    /// A signed 16-bit integer.
    Int16,
    /// A signed 32-bit integer.
    Int32,
    /// A signed 64-bit integer.
    Int64,
    /// An unsigned 8-bit integer.
    UInt8,
    /// An unsigned 16-bit integer.
    UInt16,
    /// An unsigned 32-bit integer.
    UInt32,
    /// An unsigned 64-bit integer.
    UInt64,
    /// A 16-bit floating point number.
    Float16,
    /// A 32-bit floating point number.
    Float32,
    /// A 64-bit floating point number.
    Float64,
    /// A timestamp with an optional timezone.
    ///
    /// Time is measured as a Unix epoch, counting the seconds from
    /// 00:00:00.000 on 1 January 1970, excluding leap seconds,
    /// as a 64-bit integer.
    ///
    /// The time zone is a string indicating the name of a time zone, one of:
    ///
    /// * As used in the Olson time zone database (the "tz database" or
    ///   "tzdata"), such as "America/New_York"
    /// * An absolute time zone offset of the form +XX:XX or -XX:XX, such as +07:30
    ///
    /// Timestamps with a non-empty timezone
    /// ------------------------------------
    ///
    /// If a Timestamp column has a non-empty timezone value, its epoch is
    /// 1970-01-01 00:00:00 (January 1st 1970, midnight) in the *UTC* timezone
    /// (the Unix epoch), regardless of the Timestamp's own timezone.
    ///
    /// Therefore, timestamp values with a non-empty timezone correspond to
    /// physical points in time together with some additional information about
    /// how the data was obtained and/or how to display it (the timezone).
    ///
    ///   For example, the timestamp value 0 with the timezone string "Europe/Paris"
    ///   corresponds to "January 1st 1970, 00h00" in the UTC timezone, but the
    ///   application may prefer to display it as "January 1st 1970, 01h00" in
    ///   the Europe/Paris timezone (which is the same physical point in time).
    ///
    /// One consequence is that timestamp values with a non-empty timezone
    /// can be compared and ordered directly, since they all share the same
    /// well-known point of reference (the Unix epoch).
    ///
    /// Timestamps with an unset / empty timezone
    /// -----------------------------------------
    ///
    /// If a Timestamp column has no timezone value, its epoch is
    /// 1970-01-01 00:00:00 (January 1st 1970, midnight) in an *unknown* timezone.
    ///
    /// Therefore, timestamp values without a timezone cannot be meaningfully
    /// interpreted as physical points in time, but only as calendar / clock
    /// indications ("wall clock time") in an unspecified timezone.
    ///
    ///   For example, the timestamp value 0 with an empty timezone string
    ///   corresponds to "January 1st 1970, 00h00" in an unknown timezone: there
    ///   is not enough information to interpret it as a well-defined physical
    ///   point in time.
    ///
    /// One consequence is that timestamp values without a timezone cannot
    /// be reliably compared or ordered, since they may have different points of
    /// reference.  In particular, it is *not* possible to interpret an unset
    /// or empty timezone as the same as "UTC".
    ///
    /// Conversion between timezones
    /// ----------------------------
    ///
    /// If a Timestamp column has a non-empty timezone, changing the timezone
    /// to a different non-empty value is a metadata-only operation:
    /// the timestamp values need not change as their point of reference remains
    /// the same (the Unix epoch).
    ///
    /// However, if a Timestamp column has no timezone value, changing it to a
    /// non-empty value requires to think about the desired semantics.
    /// One possibility is to assume that the original timestamp values are
    /// relative to the epoch of the timezone being set; timestamp values should
    /// then adjusted to the Unix epoch (for example, changing the timezone from
    /// empty to "Europe/Paris" would require converting the timestamp values
    /// from "Europe/Paris" to "UTC", which seems counter-intuitive but is
    /// nevertheless correct).
    Timestamp(TimeUnit, Option<String>),
    /// A 32-bit date representing the elapsed time since UNIX epoch (1970-01-01)
    /// in days (32 bits).
    Date32,
    /// A 64-bit date representing the elapsed time since UNIX epoch (1970-01-01)
    /// in milliseconds (64 bits). Values are evenly divisible by 86400000.
    Date64,
    /// A 32-bit time representing the elapsed time since midnight in the unit of `TimeUnit`.
    Time32(TimeUnit),
    /// A 64-bit time representing the elapsed time since midnight in the unit of `TimeUnit`.
    Time64(TimeUnit),
    /// Measure of elapsed time in either seconds, milliseconds, microseconds or nanoseconds.
    Duration(TimeUnit),
    /// A "calendar" interval which models types that don't necessarily
    /// have a precise duration without the context of a base timestamp (e.g.
    /// days can differ in length during day light savings time transitions).
    Interval(IntervalUnit),
    /// Opaque binary data of variable length.
    ///
    /// A single Binary array can store up to [`i32::MAX`] bytes
    /// of binary data in total
    Binary,
    /// Opaque binary data of fixed size.
    /// Enum parameter specifies the number of bytes per value.
    FixedSizeBinary(i32),
    /// Opaque binary data of variable length and 64-bit offsets.
    ///
    /// A single LargeBinary array can store up to [`i64::MAX`] bytes
    /// of binary data in total
    LargeBinary,
    /// A variable-length string in Unicode with UTF-8 encoding
    ///
    /// A single Utf8 array can store up to [`i32::MAX`] bytes
    /// of string data in total
    Utf8,
    /// A variable-length string in Unicode with UFT-8 encoding and 64-bit offsets.
    ///
    /// A single LargeUtf8 array can store up to [`i64::MAX`] bytes
    /// of string data in total
    LargeUtf8,
    /// A list of some logical data type with variable length.
    ///
    /// A single List array can store up to [`i32::MAX`] elements in total
    List(Box<Field>),
    /// A list of some logical data type with fixed length.
    FixedSizeList(Box<Field>, i32),
    /// A list of some logical data type with variable length and 64-bit offsets.
    ///
    /// A single LargeList array can store up to [`i64::MAX`] elements in total
    LargeList(Box<Field>),
    /// A nested datatype that contains a number of sub-fields.
    Struct(Vec<Field>),
    /// A nested datatype that can represent slots of differing types. Components:
    ///
    /// 1. [`Field`] for each possible child type the Union can hold
    /// 2. The corresponding `type_id` used to identify which Field
    /// 3. The type of union (Sparse or Dense)
    Union(Vec<Field>, Vec<i8>, UnionMode),
    /// A dictionary encoded array (`key_type`, `value_type`), where
    /// each array element is an index of `key_type` into an
    /// associated dictionary of `value_type`.
    ///
    /// Dictionary arrays are used to store columns of `value_type`
    /// that contain many repeated values using less memory, but with
    /// a higher CPU overhead for some operations.
    ///
    /// This type mostly used to represent low cardinality string
    /// arrays or a limited set of primitive types as integers.
    Dictionary(Box<DataType>, Box<DataType>),
    /// Exact 128-bit width decimal value with precision and scale
    ///
    /// * precision is the total number of digits
    /// * scale is the number of digits past the decimal
    ///
    /// For example the number 123.45 has precision 5 and scale 2.
    ///
    /// In certain situations, scale could be negative number. For
    /// negative scale, it is the number of padding 0 to the right
    /// of the digits.
    ///
    /// For example the number 12300 could be treated as a decimal
    /// has precision 3 and scale -2.
    Decimal128(u8, i8),
    /// Exact 256-bit width decimal value with precision and scale
    ///
    /// * precision is the total number of digits
    /// * scale is the number of digits past the decimal
    ///
    /// For example the number 123.45 has precision 5 and scale 2.
    ///
    /// In certain situations, scale could be negative number. For
    /// negative scale, it is the number of padding 0 to the right
    /// of the digits.
    ///
    /// For example the number 12300 could be treated as a decimal
    /// has precision 3 and scale -2.
    Decimal256(u8, i8),
    /// A Map is a logical nested type that is represented as
    ///
    /// `List<entries: Struct<key: K, value: V>>`
    ///
    /// The keys and values are each respectively contiguous.
    /// The key and value types are not constrained, but keys should be
    /// hashable and unique.
    /// Whether the keys are sorted can be set in the `bool` after the `Field`.
    ///
    /// In a field with Map type, the field has a child Struct field, which then
    /// has two children: key type and the second the value type. The names of the
    /// child fields may be respectively "entries", "key", and "value", but this is
    /// not enforced.
    Map(Box<Field>, bool),
    /// A run-end encoding (REE) is a variation of run-length encoding (RLE). These
    /// encodings are well-suited for representing data containing sequences of the
    /// same value, called runs. Each run is represented as a value and an integer giving
    /// the index in the array where the run ends.
    ///
    /// A run-end encoded array has no buffers by itself, but has two child arrays. The
    /// first child array, called the run ends array, holds either 16, 32, or 64-bit
    /// signed integers. The actual values of each run are held in the second child array.
    ///
    /// These child arrays are prescribed the standard names of "run_ends" and "values"
    /// respectively.
    RunEndEncoded(Box<Field>, Box<Field>),
}
```

which one do you mean?

chubei: Forgive me, I must have confused it with something else. Do you remember why we need these two types in the first place?

chloeminkyung: U128 and I128 were for Duration at the first place but it needs to support casting and stuff and e.g. timestamp - timestamp = Duration(i128) nanoseconds

dozer-api/src/grpc/types_helper.rs
chloeminkyung: yup

dozer-types/src/arrow_types/to_arrow.rs
chubei: Got it. I don't think it should be a cast. Instead should be an explicit function like `NANOSECONDS(duration)`.

chloeminkyung: Ah got it - we can discuss this on https://github.com/getdozer/dozer/pull/1374

chubei: And in your above example, `timestamp - timestamp` should return `Duration`, without going through `i128`.

chubei (approved):
Update `dozer-api/README.md` and we're good to go!

dozer-api/src/grpc/types_helper.rs
```diff
@@ -145,9 +146,9 @@ pub fn map_field_definitions(
 fn field_type_to_internal_type(typ: FieldType) -> Type {
     match typ {
         FieldType::UInt => Type::UInt,
-        FieldType::U128 => Type::UInt, // TODO: Map this correctly
+        FieldType::U128 => Type::String,
```
chubei, line 148: Sorry just saw this. We need to add `Type::U128` and `Type::I128`.

dozer-api/README.md
```diff
@@ -7,7 +7,9 @@ Following table shows how Dozer type is converted to JSON type in REST API.
 | Dozer Type | JSON Type      | Note |
 |------------|----------------|-|
 | uint       | number         | |
+| u128       | number         | |
```
chubei, line 10: This should be `string`.

dozer-api/src/generator/oapi/utils.rs
```diff
@@ -103,10 +103,12 @@ pub fn convert_cache_to_oapi_schema(
 /// Should be consistent with `field_to_json_value`.
 fn convert_cache_type_to_schema_type(field_type: dozer_types::types::FieldType) -> Type {
     match field_type {
-        FieldType::UInt | FieldType::Int => Type::Integer(IntegerType {
-            format: VariantOrUnknownOrEmpty::Item(IntegerFormat::Int64),
-            ..Default::default()
-        }),
+        FieldType::UInt | FieldType::U128 | FieldType::Int | FieldType::I128 => {
```
chubei, line 106: `U128` and `I128` should be `Type::String`.

dozer-api/README.md
chloeminkyung: fixed!!

dozer-api/src/generator/oapi/utils.rs
chloeminkyung: right

dozer-api/src/grpc/types_helper.rs
chloeminkyung: so we going to add U128 and I128 instead of String right?

dozer-api/src/grpc/types_helper.rs
```diff
@@ -86,17 +85,22 @@ fn field_to_prost_value(f: Field) -> Value {
         Field::UInt(n) => Value {
             value: Some(value::Value::UintValue(n)),
         },
+        Field::U128(n) => Value {
+            value: Some(value::Value::StringValue(n.to_string())),
```
chubei, line 89: I'm not sure about this. Should it be `Value::U128Value`?

chloeminkyung: yes yes changed

PR #1318, opened 2023-03-27
feat: make sources optional 

v3g42:
- Specifying sources is optional. Only necessary if columns are to be filtered.
- Current `table_name` logic is not changed so as to not break the samples.

Basically both options will work. 
```
app_name: local
connections:
  - config: !LocalStorage
      details:
        path: /Users/vivek/projects/dozer/rust/dozer-samples-internal/taxi-objectstore
      tables:
        - !Table
          name: trips
          prefix: data
          file_type: parquet
          extension: .parquet
    name: ny_taxi

endpoints:
  - name: trips
    path: /trips
    table_name: trips

```

```
app_name: local
connections:
  - config: !LocalStorage
      details:
        path: /Users/vivek/projects/dozer/rust/dozer-samples-internal/taxi-objectstore
      tables:
        - !Table
          name: trips
          prefix: data
          file_type: parquet
          extension: .parquet
    name: ny_taxi

sources:
  - name: trips
    table_name: trips
    connection: !Ref ny_taxi
    columns:

endpoints:
  - name: trips
    path: /trips
    table_name: trips

```


v3g42:
> No I've seen this before. https://github.com/getdozer/dozer/actions/runs/4524735735/jobs/7968758178#step:12:774
> 
> `proptest` is randomized so it can fail nondeterministically.

Ok. I ran it twice and it failed both times. One more try :)

chubei (approved):

PR #1319, opened 2023-03-27
chore: Remove `dozer-storage` dependency from `dozer-core`

### LINKED ISSUE #1205: Remove dependency from LmdbStorage from dozer-core
As not the pipeline does not have any state anymore we can remove any storage and checkpointing dependency from dozer-core

chubei:
Closes #1205


snork-alt (commented):
Shall we make SharedTransaction a generic context  so that if in the future we decide to support snapshotting it's easier to do that?

snork-alt (approved):

chubei:
> Shall we make SharedTransaction a generic context so that if in the future we decide to support snapshotting it's easier to do that?

There'll be many changes adding a generic parameter to all the processors. Snapshotting will only need the parameter in `commit` right? We can add the parameter later.

Issue #1285, opened 2023-03-22
Dozer consumes >200% of CPU even when idle


snork-alt:
Run any of the samples. When ingesting data Dozer will consume 4/500% of CPU. After ingestion is done, Dozer is till using 200% of CPU

```
33514  dozer        301.8 04:51.19 66/3   0    112   2204M+ 0B     1871M+ 33514 29284 running  *0[1]             0.00000 0.00000    501  273898     72       1032+       380+        1148367+    3235221+    771455+
```

PR #1320, opened 2023-03-27
perf: Put indexing thread pool to idle if indexes are up to date

chubei:
The indexing threads used to be looping even if there's nothing to index.

This PR fixes that. Now the indexing threads will go idle if secondary indexes are up to date. The main database commit will wake up the indexing threads.

Part of #1285


chubei:
Turns out we have to open the lmdb environment without `NO_LOCK`, because we're opening the same environment multiple times, and the readers need to be coordinates with the writer.

PR #1324, opened 2023-03-28
fix: division by zero error handling

chloeminkyung:
- [x] Division
Screenshot 2023-03-28 at 12 32 56 PM

- [x] Modulo
Screenshot 2023-03-28 at 1 01 15 PM


dozer-sql/src/pipeline/errors.rs
```diff
@@ -189,6 +189,8 @@ pub enum SqlError {
     WindowError(String),
     #[error("SQL Error: Invalid column name {0}.")]
     InvalidColumn(String),
+    #[error("SQL Error: {0} operation cannot be done due to {1}.")]
+    OperationError(String, String),
```
chubei, line 193: I'll prefer

```rust
Operation(OperationError)
```

```rust
enum OperationError {
  DivisionByZeroOrOverflow,
  ModuloByZeroOrOverflow,
}
```

chubei (commented):
Does this solve the `proptest` failure? I don't see any modification to the test.

dozer-sql/src/pipeline/errors.rs
chloeminkyung: addressed!

PR #1335, opened 2023-03-29
fix: Multimap single item removal

mediuminvader:


chubei:
What's this PR supposed to solve?

mediuminvader:
> What's this PR supposed to solve?

Currently we're using multimap to store join indexes;
if a key contains 2 records and we get 1 delete operation, we want to delete a single record from the index, as we expect 2 delete operations to delete 2 records.

PR #1336, opened 2023-03-29
fix: Multimap single item removal

mediuminvader:


chubei:
What's this PR supposed to solve?

mediuminvader:
> What's this PR supposed to solve?

Remove a single Record from a Multimap entry even if there are duplicates

chubei:
Got it. The old code removes all duplicates and that's a bug right?

chubei (approved):

PR #1342, opened 2023-03-30
feat: COALESCE() support

chloeminkyung:
- [x] implementation
- [x] prop unit test
- [x] + minor sql unit test refactor


dozer-sql/src/pipeline/expression/builder.rs
```diff
@@ -300,7 +303,13 @@ impl ExpressionBuilder {
                             fun: gft,
                             args: function_args.clone(),
                         }),
-                        Err(_err) => Err(InvalidNestedAggregationFunction(function_name)),
+                        Err(_e) => match ConditionalExpressionType::new(function_name.as_str()) {
```
mediuminvader, line 306: Is it possible to have a sequence of try and fail? rather than a nested one?

mediuminvader (changes requested):

dozer-sql/src/pipeline/expression/builder.rs
```diff
@@ -300,7 +303,13 @@ impl ExpressionBuilder {
...
+                            Ok(cet) => Ok(ConditionalExpression {
+                                fun: cet,
+                                args: function_args.clone(),
+                            }),
+                            Err(_err) => Err(InvalidNestedAggregationFunction(function_name)),
```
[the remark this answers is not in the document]
chloeminkyung: got it

dozer-sql/src/pipeline/expression/builder.rs
chloeminkyung: got it fixed

mediuminvader (approved):

PR #1355, opened 2023-03-31
fix: Retry connection to internal server

karolisg:


v3g42 (approved):

chubei:
Internal server client is removed in https://github.com/getdozer/dozer/pull/1356 😕

PR #1374, opened 2023-04-04
feat: `Duration` `Field` & `FieldType` support

chloeminkyung:
- [x] implementation
- [x] mathematical
- [x] unit tests
- [ ] will tackle on sql representation of `Duration` on separate PR (e.g. `INTERVAL '1 MINUTE'`, `DURATION '1 MINUTE'`, `"6/04/2023 10:00" + 30 MINUTES`)

```rust
pub struct DozerDuration(pub std::time::Duration, pub TimeUnit);

pub enum TimeUnit {
    Seconds,
    Milliseconds,
    Microseconds,
    Nanoseconds,
}
```

`Field` implementation
```rust
Duration(DozerDuration),
```

`FieldType` implementation
```rust
/// Duration up to nanoseconds.
Duration,
```

```rust
// Duration + Duration = Duration
// Duration - Duration = Duration
// Duration * Duration = Error
// Duration / Duration = Error
// Duration % Duration = Error

// Duration + Timestamp = Timestamp
// Duration - Timestamp = Error
// Duration * Duration = Error
// Duration / Duration = Error
// Duration % Duration = Error

// Timestamp + Duration = Timestamp
// Timestamp - Duration = Timestamp
// Timestamp * Duration = Error
// Timestamp / Duration = Error
// Timestamp % Duration = Error

// Timestamp + Timestamp = Error
// Timestamp - Timestamp = Duration (valid only when Timestamp (left) > Timestamp (right))
// Timestamp * Timestamp = Error
// Timestamp / Timestamp = Error
// Timestamp % Timestamp = Error

```


chubei:
Can you summarize:

- How is duration represented in REST API
- How is duration represented in gRPC API
- How does duration convert to and from arrow data types in gRPC connector
- How does duration convert from arrow data types in object store connector

And I remember we determined to use
```rust
pub struct DozerDuration(pub std::time::Duration, pub TimeUnit);
```

mediuminvader:
```rust
pub struct DozerDuration(pub std::time::Duration, pub TimeUnit);
```
Can you give more details on this structure? 
for my understanding Duration is assuming already a TimeUnit, how it will work?

chubei:
TimeUnit is metadata we carry from data source (arrow), and we can use this metadata when exporting to external format (arrow). It doesn't affect how we internally represent duration. I suggested to use std::time::Duration because it's more convenient and less error prone than i128.

mediuminvader:
Ok. One note for the enum, shall we consider also Minutes, Hours, Days and so on? Or do we want a different type for intervals?

chloeminkyung:
> TimeUnit is metadata we carry from data source (arrow), and we can use this metadata when exporting to external format (arrow). It doesn't affect how we internally represent duration. I suggested to use std::time::Duration because it's more convenient and less error prone than i128.

Agree on Bei's point - will proceed with `std::time::Duration`, but then i128 is not need in this case at all is it?

```
#[stable(feature = "duration", since = "1.3.0")]
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
#[cfg_attr(not(test), rustc_diagnostic_item = "Duration")]
pub struct Duration {
    secs: u64,
    nanos: Nanoseconds, // Always 0 <= nanos < NANOS_PER_SEC
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[repr(transparent)]
#[rustc_layout_scalar_valid_range_start(0)]
#[rustc_layout_scalar_valid_range_end(999_999_999)]
struct Nanoseconds(u32);
```

Correct me if I'm wrong @v3g42, @chubei, @mediuminvader we set the default as nanoseconds - but TimeUnit metadata allows us to convert into other units as well.

chubei:
> Ok. One note for the enum, shall we consider also Minutes, Hours, Days and so on? Or do we want a different type for intervals?

As long as we cover arrow time units it should be ok.

chubei:
> we set the default as nanoseconds - but TimeUnit metadata allows us to convert into other units as well.

I'm not sure what you mean by converting into other units. Per my understanding, inside Dozer all duration is of nanoseconds precision. The time unit is used when we export to external system (arrow).

For example, if time unit is seconds, we can discard the subsecond precision when exporting, resulting in u64 (or i64, I'm not sure) written to arrow.

This is mainly for telemetry purposes, where sometimes we want to use milliseconds precision, sometimes nanoseconds.

chubei:
Maybe TimeUnit is better called Precision?

chloeminkyung:
> > we set the default as nanoseconds - but TimeUnit metadata allows us to convert into other units as well.
> 
> I'm not sure what you mean by converting into other units. Per my understanding, inside Dozer all duration is of nanoseconds precision. The time unit is used when we export to external system (arrow).
> 
> For example, if time unit is seconds, we can discard the subsecond precision when exporting, resulting in u64 (or i64, I'm not sure) written to arrow.
> 
> This is mainly for telemetry purposes, where sometimes we want to use milliseconds precision, sometimes nanoseconds.

yes, this was the one that I mentioned as `converting`

chloeminkyung:
> Maybe TimeUnit is better called Precision?

both sounds right, but if we don't want to get confused, TimeUnit should be better?

chubei:
> > Maybe TimeUnit is better called Precision?
> 
> both sounds right, but if we don't want to get confused, TimeUnit should be better?

Sure. I'm ok with either

chloeminkyung:
> Can you summarize:
> 
> * How is duration represented in REST API
> * How is duration represented in gRPC API
> * How does duration convert to and from arrow data types in gRPC connector
> * How does duration convert from arrow data types in object store connector
> 
> And I remember we determined to use
> 
> ```rust
> pub struct DozerDuration(pub std::time::Duration, pub TimeUnit);
> ```

- How is duration represented in REST API
```rust
fn convert_duration_to_object(d: &DozerDuration) -> Value {
    let mut m = Map::new();
    m.insert("value".to_string(), Value::from(d.0.as_nanos().to_string())); // as_nanos() will give out u128 value which has to be represented in string as there is no support for 128 primitive value yet
    m.insert("time_unit".to_string(), Value::from(d.1.to_string()));
    Value::Object(m)
}
```

- How is duration represented in gRPC API
```rust
fn map_duration_to_prost_coord_map(d: DozerDuration) -> Value {
    Value {
        value: Some(value::Value::DurationValue(DurationType {
            value: d.0.as_nanos().to_string(),
            time_unit: d.1.to_string(),
        })),
    }
}
```

- How does duration convert to and from arrow data types in gRPC connector
```rust
FieldType::Duration => DataType::Duration(TimeUnit::Nanosecond),

FieldType::Duration => {
            let mut builder = arrow::array::DurationNanosecondArray::builder(count);
            for field in fields {
                match field {
                    Field::Duration(value) => builder.append_value(value.0.as_nanos().as_i64()),
                    Field::Null => builder.append_null(),
                    _ => panic!("Unexpected field type"),
                }
            }
            Arc::new(builder.finish())
        }
```
```rust
GrpcTypes::value::Value::DurationValue(d) => {
            let duration_type_desc = descriptor.duration_field.message.clone();
            let value_field_desc = &descriptor.duration_field.value;
            let time_unit_field_desc = &descriptor.duration_field.time_unit;
            let mut duration = DynamicMessage::new(duration_type_desc);
            duration.set_field(value_field_desc, prost_reflect::Value::String(d.value));
            duration.set_field(
                time_unit_field_desc,
                prost_reflect::Value::String(d.time_unit),
            );
            Value::Message(duration)
        }
```

mediuminvader (approved):

v3g42 (approved):
Looks good to me.

PR #1381, opened 2023-04-05
feature: Read object store files in parallel

### LINKED ISSUE #934: Parallelize the ingestion of ObjectStore sources
If a single connector has multiple sources, readingof sources is not parallelized

karolisg:
Closes #934

### DIFF
--- a/dozer-ingestion/src/connectors/object_store/table_reader.rs
+++ b/dozer-ingestion/src/connectors/object_store/table_reader.rs
@@ -159,10 +159,7 @@ impl<T: DozerObjectStore> Reader<T> for TableReader<T> {
                 .await;
                 if let Err(e) = result {
                     sender.send(Err(e)).unwrap();
-                    return Err(ObjectStoreConnectorError::DataReadError);
                 }
-
-                Ok(())
             });
         }
 
--- a/dozer-ingestion/src/connectors/object_store/table_reader.rs
+++ b/dozer-ingestion/src/connectors/object_store/table_reader.rs
@@ -2,13 +2,14 @@ use crate::connectors::object_store::adapters::DozerObjectStore;
 use crate::connectors::object_store::helper::map_listing_options;
 use crate::connectors::object_store::schema_helper::map_value_to_dozer_field;
 use crate::connectors::TableInfo;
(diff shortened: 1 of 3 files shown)


dozer-ingestion/src/errors.rs
```diff
@@ -402,6 +402,9 @@ pub enum ObjectStoreConnectorError {
 
     #[error(transparent)]
     IngestorError(#[from] IngestorError),
+
+    #[error("Failed to send message on data read channel")]
+    DataReadError,
```
chubei, line 407: Include error source?

dozer-ingestion/src/connectors/object_store/table_reader.rs
```diff
@@ -143,21 +144,48 @@ impl<T: DozerObjectStore> Reader<T> for TableReader<T> {
...
             );
 
-            rt.block_on(Self::read(
-                id as u32,
-                ctx,
-                table_path,
-                listing_options,
-                ingestor,
-                table,
-            ))?;
```
chubei, line 153: We shouldn't start many tokio runtimes. Every table read is an asynchronous task, one runtime should be enough to parallelize them.

chubei (changes requested):

dozer-ingestion/src/connectors/object_store/table_reader.rs
```diff
@@ -147,27 +147,30 @@ impl<T: DozerObjectStore> Reader<T> for TableReader<T> {
...
                     listing_options,
                     &t,
                     sender.clone(),
-                ));
-
+                )
+                .await;
                 if let Err(e) = result {
                     sender.send(Err(e)).unwrap();
+                    return Err(ObjectStoreConnectorError::DataReadError);
```
chubei, line 162: What's the purpose of returning error here?

chubei (approved):

dozer-ingestion/src/connectors/object_store/table_reader.rs
karolisg: Right, there is no purpose. Removed it in last commit.

chubei: Also remove `DataReadError`?

PR #1398, opened 2023-04-07
feature: Implement parse pipeline Admin API

karolisg:


dozer-admin/src/services/application_service.rs
```diff
@@ -80,6 +92,113 @@ impl AppService {
...
+                                columns,
+                                table_name: endpoint_info.table_name.clone(),
+                            },
+                        );
+                    }
+                }
+            }
+            Ok((calculated_sources.original_sources, transformed_sources))
+        })
+        .join()
```
chubei, line 188: Spawning a thread and immediately joining it looks unnecessary

chubei (changes requested):

PR #1403, opened 2023-04-09
feat: implement log crate + js bindings

v3g42:


chubei:
Included in #1420 and #1424

PR #1413, opened 2023-04-11
fix: Use `tokio::time::sleep` instead of `std::thread::sleep`

chubei:
Otherwise one of the tokio worker threads is blocked and can't be used for other tasks.

Also, tokio runtime will block forever when shutdown because we never yield.

### DIFF
--- a/dozer-api/examples/reader.rs
+++ b/dozer-api/examples/reader.rs
@@ -1,7 +1,6 @@
 use std::{env, path::Path};
 
 use dozer_api::LogReader;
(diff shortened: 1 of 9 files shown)


v3g42 (approved):

chubei:
# `main`

```text
cargo run --release --example reader  0.43s user 0.12s system 66% cpu 0.834 total
```

# PR

```text
cargo run --release --example reader  0.43s user 0.16s system 66% cpu 0.886 total
```

Minor performance degrade.

PR #1441, opened 2023-04-18
feat: allow disabling of grpc & rest apis via config

universalmind303:
There didn't seem to be any way of disabling either the rest or gRPC apis, this PR adds a new `enabled` flag to both `rest` and `grpc`. If `enabled: false`, the service will not be started at all.

### DIFF
--- a/dozer-orchestrator/src/simple/orchestrator.rs
+++ b/dozer-orchestrator/src/simple/orchestrator.rs
@@ -102,29 +102,37 @@ impl Orchestrator for SimpleOrchestrator {
 
             // Initialize API Server
             let rest_config = get_rest_config(self.config.to_owned());
-            let security = get_api_security_config(self.config.to_owned());
-            let cache_endpoints_for_rest = cache_endpoints.clone();
-            let shutdown_for_rest = shutdown.create_shutdown_future();
-            let rest_handle = tokio::spawn(async move {
-                let api_server = rest::ApiServer::new(rest_config, security);
-                api_server
-                    .run(cache_endpoints_for_rest, shutdown_for_rest)
-                    .await
-                    .map_err(OrchestrationError::ApiServerFailed)
-            });
+            let rest_handle = if rest_config.enabled {
+                let security = get_api_security_config(self.config.to_owned());
+                let cache_endpoints_for_rest = cache_endpoints.clone();
+                let shutdown_for_rest = shutdown.create_shutdown_future();
+                tokio::spawn(async move {
+                    let api_server = rest::ApiServer::new(rest_config, security);
+                    api_server
+                        .run(cache_endpoints_for_rest, shutdown_for_rest)
+                        .await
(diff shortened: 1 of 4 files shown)


mediuminvader:
Hi @universalmind303 thanks for the PR this is a good idea, but I would discuss a bit about the behavior.

My proposal is to make API disabled by default, and enabled by default if rest and/or grpc item is in the configuration, without need specify if it's enable.

For instance, this configuration will enable grpc and disable rest:
```
app_name: whatever
api:
  grpc:
    port: 9090
```
 this configuration will enable rest and disable grpc:
```
app_name: whatever
api:
  rest:
    port: 8080
```
for convenience we can use `enable: false` to temporary disable a service, without having to write/delete the block:
```
app_name: whatever
api:
  rest:
    enable: false
    port: 8080
  grpc:
    port: 9090
```
an option could be to enable REST by default, but that will breaks the convention.

This is just a proposal, any comment @getdozer/dozer-dev ?

chubei:
I prefer the mechanism implemented in this PR because otherwise we'll have to modify all samples to enable apis.

mediuminvader:
> I prefer the mechanism implemented in this PR because otherwise we'll have to modify all samples to enable apis.

good point

v3g42:
> I prefer the mechanism implemented in this PR because otherwise we'll have to modify all samples to enable apis.
> 
> They don't have api section right now and are using the default ports.

I agree. This is a non-breaking behavior and feels intuitive. 
Given that we are able to run `app` and `api` portions separately, in the future we should have ability to combine multiple YAML files as well. `app` portions would not need the `api` block and vice versa.

mediuminvader:
> > I prefer the mechanism implemented in this PR because otherwise we'll have to modify all samples to enable apis.
> > They don't have api section right now and are using the default ports.
> 
> I agree. This is a non-breaking behavior and feels intuitive. Given that we are able to run `app` and `api` portions separately, in the future we should have ability to combine multiple YAML files as well. `app` portions would not need the `api` block and vice versa.

Makes sense, let's stick to the PR approach.

abhishekmishragithub:
> I prefer the mechanism implemented in this PR because otherwise we'll have to modify all samples to enable apis.
> 
> They don't have api section right now and are using the default ports.

+1
I agree. The current PR doesn’t break the change and ruining samples. If we implement what @mediuminvader suggested then we need update the references everywhere.
We should go with current approach l.

universalmind303:
I think I have all of the tests passing now. I saw that this was tagged with `doc-update-needed`. I couldn't find where those docs are. Are they in a separate repo?

mediuminvader (approved):

chubei:
Hi @universalmind303 the label is just a reminder for updating the docs. Don't worry about it.

You still need to sign your commits before this PR can be merged.

chubei:
Just FYI we've lifted the requirement of signing commits, so future contribution can be easier.

PR #1450, opened 2023-04-20
chore: Add get status method to admin proto service

karolisg:


chubei:
Missing implementation?

karolisg:
> Missing implementation?

We have remove dozer-admin from this repo, so implementation is in the different repo now.

PR #1458, opened 2023-04-21
fix: uuid postgres type support

chloeminkyung:
(cherry picked from commit 82551d116a68b26a7b9a1088971fd220745b4eab)

### DIFF
--- a/dozer-ingestion/src/connectors/postgres/helper.rs
+++ b/dozer-ingestion/src/connectors/postgres/helper.rs
@@ -14,6 +14,7 @@ use rust_decimal::Decimal;
 use std::error::Error;
 use std::vec;
 use tokio_postgres::{Column, Row};
+use uuid::Uuid;
 
 use dozer_types::geo::Point as GeoPoint;
 
@@ -38,6 +39,7 @@ pub fn postgres_type_to_field(
                 Type::TEXT | Type::VARCHAR | Type::CHAR | Type::BPCHAR => {
                     Ok(Field::String(String::from_utf8(v.to_vec()).unwrap()))
                 }
+                Type::UUID => Ok(Field::String(String::from_utf8(v.to_vec()).unwrap())),
                 Type::BYTEA => Ok(Field::Binary(v.to_vec())),
                 Type::NUMERIC => Ok(Field::Decimal(
                     Decimal::from_f64(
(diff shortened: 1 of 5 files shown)


chubei (commented):
Can you extend the tests in https://github.com/getdozer/dozer/blob/main/dozer-ingestion/tests/test_suite/connectors/sql.rs to include `uuid`?

dozer-ingestion/src/connectors/postgres/helper.rs
```diff
@@ -153,6 +157,11 @@ pub fn value_to_field(
             })
         }
         &Type::POINT => convert_row_value_to_field!(row, idx, GeoPoint),
+        // &Type::UUID => convert_row_value_to_field!(row, idx, Uuid),
+        &Type::UUID => {
```
chubei, line 161: @karolisg is it possible to refactor the code so we don't duplicate the conversion code? Now we have `value_to_field` and `postgres_type_to_field` doing the same thing.

chloeminkyung: currently, 

```rust
pub fn value_to_field(
    row: &Row,
    idx: usize,
    col_type: &Type,
) -> Result<Field, PostgresSchemaError> {
```

```rust 
pub fn postgres_type_to_field(
    value: Option<&Bytes>,
    column: &TableColumn,
) -> Result<Field, PostgresSchemaError> {
```

I was thinking replacing `postgres_type_to_field` with `value_to_field` but we don't have access to `Row` where we are calling `postgres_type_to_field()` at the moment, this might be a bigger refactor than we think.

chubei: > this might be a bigger refactor than we think

True. And neither of us is familiar enough with that part of code, so I was asking @karolisg. It won't be part of this PR.

dozer-ingestion/tests/test_suite/connectors/sql.rs
```diff
@@ -39,7 +39,8 @@ pub fn create_table_with_all_supported_data_types(table_name: &str) -> String {
             date DATE NOT NULL,
             date_null DATE,
             point POINT NOT NULL,
-            point_null POINT
+            point_null POINT,
+            uuid UUID NOT NULL
```
chubei, line 43: Let's also include the null uuid case

dozer-ingestion/tests/test_suite/connectors/sql.rs
```diff
@@ -39,7 +39,8 @@ pub fn create_table_with_all_supported_data_types(table_name: &str) -> String {
             date DATE NOT NULL,
             date_null DATE,
             point POINT NOT NULL,
-            point_null POINT
+            point_null POINT,
+            uuid UUID
```
chubei, line 43: Let's include both the "null" and "NOT NULL" case

chubei (approved):

PR #1459, opened 2023-04-21
chore: Change error message for `CustomTypeNotSupported`

chubei:


v3g42 (approved):

dozer-ingestion/src/errors.rs
```diff
@@ -218,7 +218,7 @@ pub enum PostgresSchemaError {
     #[error("Column type {0} not supported")]
     ColumnTypeNotSupported(String),
 
-    #[error("CustomTypeNotSupported")]
+    #[error("Custom type is not supported yet. Join our Discord at https://discord.com/invite/3eWXBgJaEQ - we're here to help with your use case!")]
```
v3g42, line 221: Shoudl we include the type name ? 

Also if there are multiple types that are not supported. Are we giving them errors for the first type or all of them together ?

chubei: Right we should include. I believe we quit on first error, @karolisg can you confirm?

PR #1465, opened 2023-04-22
Fix cli init generation

nurikk:
This pr fixes config file generated using init command.

Now instead of (which causes cache files to create inside working directory)
```yaml
home_dir: ./dozer
cache_dir: ''
```

it generates (./dozer instead of ./dozer as well as cache folder located inside home_dir)

```yaml
home_dir: ./.dozer
cache_dir: ./.dozer/cache
```


chubei (approved):
Thanks!

PR #1500, opened 2023-04-26
feat(ingestion): properly support postgres with SSL OFF

universalmind303:
currently if you try to connect to certain postgres instances, even with SSL off, you will get the following error message

```
ERROR Failed to connect to postgres with the specified configuration. db error: FATAL: no pg_hba.conf entry for host "_", user "_", database "_", SSL off
```

Now if ssl is disabled (which it always is) it will not verify.  

As a follow up, I think we should also expose some configuration params on the `PostgresConfig` struct for supporting various ssl modes. 

The implementation was largely inspired by the [connectorx](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/postgres/connection.rs) postgres connector. 

## Additional notes.
I didn't really know how to go about creating a test for this specific case. I manually tested it, and the above issue is gone when connecting to my pg instance. _(and it's pretty much the same as the connectorx implementation)._

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2434,7 +2434,9 @@ dependencies = [
  "kafka",
  "object_store",
  "odbc",
+ "openssl",
  "parquet 33.0.0",
+ "postgres-openssl",
  "postgres-protocol",
  "postgres-types",
  "prost",
@@ -4985,6 +4987,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "postgres-openssl"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1de0ea6504e07ca78355a6fb88ad0f36cafe9e696cbc6717f16a207f3a60be72"
+dependencies = [
+ "futures",
+ "openssl",
+ "tokio",
+ "tokio-openssl",
+ "tokio-postgres",
+]
+
+[[package]]
 name = "postgres-protocol"
 version = "0.6.4"
 source = "git+https://github.com/getdozer/rust-postgres#0ecaeddecbba32600e8ecc2026ebdc663170551e"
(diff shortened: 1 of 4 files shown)


chubei:
Hi @universalmind303, thank you for identifying and fixing this issue!

While we definitely need the fix, we're hesitant to introduce `openssl-sys` as the dependency of our release. This created deployment issues in the past and we've put all connectors requiring `openssl-sys` behind feature flags and not enable them by default. The plan is to migrate them to `rustls` one by one and include them in the release.

Long story short, we'll investigate if it's possible to connect to postgres with `rustls`, if not, we'll need this openssl dependency behind a feature flag.

On another note, maybe you can change your postgres `pg_hba.conf` to allow client to connect to it without verifying ssl? Here's an example https://github.com/getdozer/dozer/blob/main/dozer-tests/src/e2e_tests/connections/bookings_conn/pg_hba.conf.

chubei:
Hi @universalmind303, I opened #1502 as an attempt to use `rustls` for ssl verification.

I wasn't able to reproduce the issue you experienced so am not sure if it fixes the problem.

Would you kindly share your database configuration and dozer configuration? Or test with #1502 if that's not possible?

universalmind303:
> While we definitely need the fix, we're hesitant to introduce openssl-sys as a dependency of our release.

i definitely understand the hesitation around adding the `openssl` crate.  I've had a number of issues with that dependency as well. 

Unfortunately, It looks like this is still trying to verify the CA

```
Sending fatal alert UnknownCA 
```
I don't have access to the conf for this specific DB. I'll do some digging & see if I can come up with a reproducable `pg_hba.conf`

chubei:
Hi @universalmind303 we haven't heard from you for a while. I'll close this PR for now. Feel free to reopen if you have anything that you want to share.

PR #1502, opened 2023-04-27
fix: Use `rustls` for tls support when connecting to postgres

chubei:


universalmind303:
> Hi @universalmind303, continuing our discussion in [#1500 (comment)](https://github.com/getdozer/dozer/pull/1500#issuecomment-1525835252), I realized I didn't add the root certificates from the system, which may be why it's reporting `UnknownCA`.
> 
> I pushed another commit that loads system root certificates using [`rustls-native-certs`](https://github.com/rustls/rustls-native-certs). Is it possible for you to check if it fixes the problem?
> 
> And if your postgres instance certificate is not trusted by your client system, you can set `SSL_CERT_FILE` env var accroding to `rustls-native-certs` documentation.

This appears to still have the same issue

```
2023-05-09T14:31:43.535359Z  WARN Sending fatal alert UnknownCA    
2023-05-09T14:31:43.536817Z ERROR Failed to connect to postgres with the specified configuration. error performing TLS handshake: invalid peer certificate: UnknownIssuer
```

I expect this to be an issue with the postgres instance. It is a managed [Heroku Postgres](https://devcenter.heroku.com/categories/heroku-postgres) instance. Which doesn't expose much in terms of configurability. Heroku also doesn't support REPLICATION, so it may not be worth investigating further.

chubei:
Thank you! I'll try to start a Heroku instance to test - not to try to run Dozer against it, just to resolve the SSL issue.

universalmind303:
@chubei I think this ended up being something with the certs for my pg instance. I was able to get a working connection without openssl by using `native_tls` option of [accept_invalid_certs](https://docs.rs/native-tls/0.2.11/native_tls/struct.TlsConnectorBuilder.html#method.danger_accept_invalid_certs). 

For flexibility, maybe it makes sense to allow users to specify  [accept_invalid_certs](https://docs.rs/native-tls/0.2.11/native_tls/struct.TlsConnectorBuilder.html#method.danger_accept_invalid_certs) and [accept_invalid_hostnames](https://docs.rs/native-tls/0.2.11/native_tls/struct.TlsConnectorBuilder.html#method.danger_accept_invalid_hostnames) when creating a connection?

chubei:
Thank you for letting us know! We'll continue to investigate if rustls is a possible solution, with similar options so it's easier for people to test!

chloeminkyung (approved):

chubei:
@universalmind303 We've tested with a https://neon.tech/ postgres instance and using the code in this PR the connection works, so we're going to merge this one in favor of #1500.

Meanwhile, do you still want the `accept_invalid_certs` option? I believe it's also possible in `rustls`, using [`DangerousClientConfig::set_certificate_verifier`](https://docs.rs/rustls/latest/rustls/client/struct.DangerousClientConfig.html#method.set_certificate_verifier).

PR #1513, opened 2023-04-28
chore: Add cloud feature flag

karolisg:


dozer-orchestrator/src/cli/types.rs
```diff
@@ -2,6 +2,7 @@ use clap::{Args, Parser, Subcommand};
```
chubei, line 2: We can put `struct Cloud` and relavant `use` and `struct` in a module and reduce two `#[cfg(feature = "cloud")]`

chubei (approved):

PR #1514, opened 2023-04-28
chore: json arrow conversion support

chloeminkyung:
`to_arrow()`: `JSON` -> `DataType::Utf8` with metadata containing {"logical_type" : "Json"}

`from_arrow()`: `DataType::Utf8` -> `JSON`  with checking metadata containing {"logical_type" : "Json"}


dozer-types/src/arrow_types/from_arrow.rs
```diff
@@ -165,7 +169,22 @@ pub fn map_arrow_to_dozer_type(dt: &DataType) -> Result<FieldType, FromArrowErro
         DataType::Binary | DataType::FixedSizeBinary(_) | DataType::LargeBinary => {
             Ok(FieldType::Binary)
         }
-        DataType::Utf8 => Ok(FieldType::String),
+        DataType::Utf8 => {
```
chubei, line 172: Utf8 handling in `map_value_to_dozer_field` needs to be handled accordingly.

Given that this is a buggy implementation but no test fails, I think we need round trip unit test: `Record` -> `RecordBatch` -> `Record`. `Record` should contain all kinds of fields.

chubei (changes requested):

dozer-types/src/arrow_types/from_arrow.rs
chloeminkyung: valid points.

1. But for `map_value_to_dozer_field` there is no way for us to get the metadata to check whether it's for `Json` or `Utf8`

```rust
DataType::Utf8 => {
            let array: Option<&serde_json::Value> =
                column.as_any().downcast_ref::<serde_json::Value>();
            if let Some(r) = array {
                let s: DozerField = if r.is_null() {
                    DozerField::Null
                } else {
                    DozerField::Json(
                        serde_json_to_json_value(r.clone())
                            .map_err(|_| FieldTypeNotSupported(column_name.to_string()))
                            .unwrap(),
                    )
                };
                Ok(s)
            } else {
                Ok(DozerField::Null)
            }
        }
```
would this be enough?

2. for the `field_test_cases` we already have the `Json` type, `test_field_serialize_roundtrip()` and this has passed. I don't get what you mean `round trip unit test: Record -> RecordBatch -> Record. Record should contain all kinds of fields.`?

chloeminkyung: added!

dozer-types/src/arrow_types/from_arrow.rs
```diff
@@ -142,12 +145,31 @@ macro_rules! make_text {
     }};
 }
 
+macro_rules! make_json {
```
chubei, line 148: I feel these macros are bad practice. What do you feel during writing code? Should we change it to a function?

dozer-types/src/arrow_types/from_arrow.rs
```diff
@@ -184,7 +208,22 @@ pub fn map_arrow_to_dozer_type(dt: &DataType) -> Result<FieldType, FromArrowErro
         DataType::Binary | DataType::FixedSizeBinary(_) | DataType::LargeBinary => {
             Ok(FieldType::Binary)
         }
-        DataType::Utf8 => Ok(FieldType::String),
+        DataType::Utf8 => {
+            if m.contains_key(LOGICAL_TYPE_KEY) {
```
chubei, line 212: No need to check `contains_key`

dozer-types/src/arrow_types/from_arrow.rs
chubei: I think we need a round trip unit test: `Record` -> `RecordBatch` -> `Record`

chubei (changes requested):

dozer-types/src/arrow_types/from_arrow.rs
```diff
@@ -257,7 +296,22 @@ pub fn map_value_to_dozer_field(
         DataType::Binary => make_binary!(array::BinaryArray, column, row),
         DataType::FixedSizeBinary(_) => make_binary!(array::FixedSizeBinaryArray, column, row),
         DataType::LargeBinary => make_binary!(array::LargeBinaryArray, column, row),
-        DataType::Utf8 => make_from!(array::StringArray, column, row),
+        DataType::Utf8 => {
+            if metadata.contains_key(LOGICAL_TYPE_KEY) {
```
chubei, line 300: No need to check `contains_key`

dozer-types/src/arrow_types/from_arrow.rs
chloeminkyung: right - agree on changing to function - but then it will exactly the same way that we are doing now but in single function with `match` to handle different type cases

chubei: I mean, to start with, we only change `make_json` to a function, leaving other parts untouched.

dozer-types/src/arrow_types/from_arrow.rs
chloeminkyung: changed!

dozer-types/src/arrow_types/from_arrow.rs
chloeminkyung: changed

dozer-types/src/arrow_types/from_arrow.rs
chloeminkyung: sure

dozer-ingestion/src/connectors/grpc/adapter/arrow.rs
```diff
@@ -126,7 +126,7 @@ pub fn handle_message(
     Ok(())
 }
 
-fn map_record_batch(
+pub fn map_record_batch(
```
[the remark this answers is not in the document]
chloeminkyung: sure

dozer-ingestion/tests/test_suite/connectors/arrow.rs
```diff
@@ -260,9 +265,15 @@ pub fn record_batch_with_all_supported_data_types() -> arrow::record_batch::Reco
...
         Arc::new(LargeStringArray::from_iter_values(["", "1"])),
         Arc::new(LargeStringArray::from_iter([Some(""), None])),
+        Arc::new(StringArray::from_iter_values(["[1, 2, 3]", "1"])),
+        Arc::new(StringArray::from_iter([Some("[1, 2, 3]"), None])),
     ];
 
-    arrow::record_batch::RecordBatch::try_new(Arc::new(schema), columns)
+    let metadata = HashMap::from([(
+        DOZER_SCHEMA_KEY.to_string(),
+        serde_json::to_string(&schema).expect("Schema can always be serialized as JSON"),
```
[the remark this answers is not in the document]
chubei: For connector test let's test the case without dozer schema metadata, which should be used more by the users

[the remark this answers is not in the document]
chloeminkyung: sure

dozer-ingestion/tests/test_suite/connectors/object_store/local_storage.rs
```diff
@@ -100,3 +102,59 @@ fn create_connector(
...
+        SourceDefinition::Dynamic,
+    );
+
+    let schema: Schema = Schema::empty()
+        .field(id, false)
+        .field(name, false)
+        .field(json, false)
+        .clone();
+
+    let records: Vec<Field> = vec![
```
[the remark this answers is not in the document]
chloeminkyung: got it but for this test we need to disregards 128 bit datatypes as they cannot be converted

dozer-types/src/arrow_types/tests.rs
```diff
@@ -1,18 +1,65 @@
+use crate::arrow_types::from_arrow::map_schema_to_dozer;
+use crate::arrow_types::to_arrow::DOZER_SCHEMA_KEY;
+use crate::types::{FieldDefinition, FieldType, Schema as DozerSchema, SourceDefinition};
 use arrow::datatypes::{self as arrow_types};
+
 use std::collections::HashMap;
 
-use crate::arrow_types::from_arrow::map_schema_to_dozer;
 #[test]
```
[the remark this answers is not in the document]
chloeminkyung: done!

dozer-ingestion/tests/test_suite/connectors/object_store/local_storage.rs
```diff
@@ -100,3 +102,59 @@ fn create_connector(
 
     (temp_dir, connector)
 }
+
+#[test]
```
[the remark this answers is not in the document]
chloeminkyung: done!

dozer-types/src/types/field.rs
```diff
@@ -769,6 +772,240 @@ pub fn field_test_cases() -> impl Iterator<Item = Field> {
     .into_iter()
 }
 
+pub fn arrow_field_test_cases() -> impl Iterator<Item = Field> {
```
chubei, line 775: Let's implement this function with using `Iterator::filter` to remove the fields that are not supported by arrow, instead of copy the fields.

dozer-types/src/types/field.rs
```diff
@@ -769,6 +772,240 @@ pub fn field_test_cases() -> impl Iterator<Item = Field> {
...
+            JsonValue::Number(OrderedFloat(111_f64)),
+            JsonValue::Number(OrderedFloat(111_f64)),
+            JsonValue::Number(OrderedFloat(34_f64)),
+        ])),
+        Field::Null,
+    ]
+    .into_iter()
+}
+
+pub fn schema_test_cases() -> Schema {
```
chubei, line 815: Rename to `arrow_field_test_cases_schema`?

chubei (commented):
Looks great! Still have a question. Could you elaborate?

dozer-types/src/arrow_types/from_arrow.rs
```diff
@@ -142,8 +146,41 @@ macro_rules! make_text {
...
+fn make_json(column: &ArrayRef, row: &usize) -> Result<DozerField, FromArrowError> {
+    let array = column.as_any().downcast_ref::<array::StringArray>();
+
+    return if let Some(r) = array {
+        let s: DozerField = if r.is_null(*row) {
+            DozerField::Null
+        } else {
+            match JsonValue::from_str(r.value(*row)) {
+                Ok(j) => DozerField::Json(j),
+                Err(_) => DozerField::from(r.value(*row)),
```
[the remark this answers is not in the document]
chloeminkyung: `r.value(*row)` is an element in the `StringArray` so it will be in string value. If that value is not able to be converted to `JsonValue` by function `from_str` we take the element `r.value(*row)` as a string itself and create string `Field` value using that line by doing `DozerField::from(r.value(*row))`

dozer-types/src/types/field.rs
chloeminkyung: sure

dozer-types/src/arrow_types/from_arrow.rs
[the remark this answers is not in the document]
chubei: Do you think it's reasonable if we return the deserialization error here?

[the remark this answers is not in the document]
chloeminkyung: hmm, no. Actually, instead i think it would be better if we just do `DozerField::Json(JsonValue::String(r.value(*row)))`. what do you think?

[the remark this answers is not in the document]
chloeminkyung: let me add `JsonDeserializationError` in `FromArrowError` then

chubei (approved):
🎉

PR #1518, opened 2023-04-28
feat: Last version check

mediuminvader:


dozer-orchestrator/src/main.rs
```diff
@@ -31,6 +35,66 @@ fn render_logo() {
...
+async fn check_update() {
+    use std::println as info;
+    const VERSION: &str = env!("CARGO_PKG_VERSION");
+    const ARCH: &str = std::env::consts::ARCH;
+    const OS: &str = std::env::consts::OS;
+
+    loop {
+        info!("Checking for updates...");
+        let request_url = format!(
+            "https://metadata.dev.getdozer.io/?version={}&build={}&os={}",
```
v3g42, line 71: I think these are better passed as query params to get instead of us constructing message directly ?

dozer-orchestrator/src/main.rs
```diff
@@ -31,6 +35,66 @@ fn render_logo() {
...
+            VERSION, ARCH, OS
+        );
+        let response = reqwest::get(&request_url).await;
+        match response {
+            Ok(r) => {
+                let package: DozerPackage = r.json().await.unwrap();
+                let current = version_to_vector(VERSION);
+                let remote = version_to_vector(&package.latest_version);
+
+                if compare_versions(remote, current) {
```
v3g42, line 81: Is it printing this message on every iteration? We should only print on first I believe

dozer-orchestrator/src/main.rs
mediuminvader: Makes sense, was just referring to the rust cookbook, fixed.

dozer-orchestrator/src/main.rs
mediuminvader: We print and stop checking?

mediuminvader: makes sense

v3g42 (approved):

PR #1526, opened 2023-05-03
feat: `JSON_VALUE()` & `JSON_QUERY()` sql support

chloeminkyung:
feat: `JSON_VALUE()` & `JSON_QUERY()` sql support

[`JSON_VALUE( expression , path )`](https://learn.microsoft.com/en-us/sql/relational-databases/json/validate-query-and-change-json-data-with-built-in-functions-sql-server?view=sql-server-ver16#VALUE) extracts a scalar value from a JSON string.
[`JSON_QUERY( expression [ , path ] )`](https://learn.microsoft.com/en-us/sql/relational-databases/json/validate-query-and-change-json-data-with-built-in-functions-sql-server?view=sql-server-ver16#QUERY) extracts an object or an array from a JSON string. Default path for query function is $ which means the original json.

```json
{
  "test_jsonb": {"bar": {"1": 1, "2": 2, "3": 3}, "baz": null, "foo": [1, 2, 3]},
  "test_json": [{"digit": 30, "letter": "A"}, {"digit": 31, "letter": "B"}]
}
```

```rust
select 
    JSON_QUERY(test_jsonb,'$.bar') AS bar,
    JSON_VALUE(test_jsonb,'$.baz') AS baz,
    JSON_QUERY(test_jsonb,'$.foo') AS foo,
    JSON_VALUE(test_json,'$[0].letter') AS first_letter,
    JSON_QUERY(test_json) AS all
  into test_uuid_test
  from test_uuid;
  
  select
    JSON_QUERY(test_json,'$[*].digit') AS all_digits
  into test_uuid_res
  from test_uuid;
```

Result

<details>

```json
"bar": {
    "struct_value": {
        "fields": [
            {
                "key": "1",
                "value": {
                    "number_value": 1
                }
            },
            {
                "key": "2",
                "value": {
                    "number_value": 2
                }
            },
            {
                "key": "3",
                "value": {
                    "number_value": 3
                }
            }
        ]
    }
},
"baz": {
    "null_value": "NULL_VALUE"
},
"foo": {
    "list_value": {
        "values": [
            {
                "number_value": 1
            },
            {
                "number_value": 2
            },
            {
                "number_value": 3
            }
        ]
    }
},
"first_letter": {
    "list_value": {
        "values": [
            {
                "string_value": "A"
            }
        ]
    }
},
"all": {
    "list_value": {
        "values": [
            {
                "list_value": {
                    "values": [
                        {
                            "struct_value": {
                                "fields": [
                                    {
                                        "key": "letter",
                                        "value": {
                                            "string_value": "A"
                                        }
                                    },
                                    {
                                        "key": "digit",
                                        "value": {
                                            "number_value": 30
                                        }
                                    }
                                ]
                            }
                        },
                        {
                            "struct_value": {
                                "fields": [
                                    {
                                        "key": "letter",
                                        "value": {
                                            "string_value": "B"
                                        }
                                    },
                                    {
                                        "key": "digit",
                                        "value": {
                                            "number_value": 31
                                        }
                                    }
                                ]
                            }
                        }
                    ]
                }
            }
        ]
    }
}
```

```json
"all_digits": {
    "list_value": {
        "values": [
            {
                "number_value": 30
            },
            {
                "number_value": 31
            }
        ]
    }
}
```

</details>


mediuminvader:
Is there a way to remove the `struct_value`, `number_value`, etc. from the output?

chloeminkyung:
> Is there a way to remove the `struct_value`, `number_value`, etc. from the output?

why do we need to remove?

chloeminkyung:
> > > Is there a way to remove the `struct_value`, `number_value`, etc. from the output?
> > 
> > 
> > why do we need to remove?
> 
> As a Json output I expect something like this:
> 
> ```json
> {
>   "bar": {
>     "1": 1, 
>     "2": 2, 
>     "3": 3
>   }
> }
> ```
> 
> is there a reason why we introduce all those extra fields?

Oh that's how gRPC represents the json value. REST result will be like what you're saying

snork-alt:
@chloeminkyung 
- what happens if I call `JSON_QUERY()` and the return value from the function happens to be a scalar?
- when iIinvoke JSON_VALUE() what is the returned scalar type? Shouldn't I specify the return type in the function for the schema to be calculated ?

chloeminkyung:
> @chloeminkyung
> 
> * what happens if I call `JSON_QUERY()` and the return value from the function happens to be a scalar?
> * when iIinvoke JSON_VALUE() what is the returned scalar type? Shouldn't I specify the return type in the function for the schema to be calculated ?

First, `JSON_QUERY()` is inclusive for `JSON_VALUE()`.
If we use `JSON_QUERY()` for scalar value, we still get the value as scalar value and we can set `lax mode` or `strict mode`. By default, it's `lax mode`, on `strict mode`, it returns `Err`.

e.g. 
```json
{
    "info": {
        "type": 1,
        "address": {
            "town": "Cheltenham",
            "county": "Gloucestershire",
            "country": "England"
        },
        "tags": ["Sport", "Water polo"]
    },
    "type": "Basic"
}
```
```sql
SELECT JSON_QUERY(jsonInfo,'$.type') FROM users
```
Result: `"Basic"`


But if we do the other way round, 
```sql
SELECT JSON_VALUE(jsonInfo,'$.info.address.tags') FROM users
```
Result: `Null`


Second, the case where we want to export a field out of json like following,

```sql
SELECT CAST(JSON_VALUE(jsonInfo, '$[0].digit') AS FLOAT) FROM users
```

then we need to use `JSON_VALUE()` for strict field typing and casting! :)

snork-alt (approved):
Add a README in the parser root folder saying 

This parser is forked from ...

PR #1537, opened 2023-05-05
Add workflow step to push Dozer image to Dockerhub

abhishekmishragithub:
Add the docker hub push workflow step in `release.yaml` workflow file. 

**Note:**
   - Add `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` as a secret to the repo settings.
   - Added a condition to push to docker hub only when there is a release.


.github/workflows/release.yaml
```diff
@@ -427,6 +429,22 @@ jobs:
...
+      - name: Build and push Docker image to Docker Hub
+        if: (needs.prepare.outputs.prerelease == 'false') && (matrix.os == 'ubuntu-20-16-cores')
+        env:
+          IMAGE_TAG: ${{ needs.prepare.outputs.version }}
+        uses: docker/build-push-action@v2
+        with:
+          context: ci/Dockerfile
+          push: true
+          tags: $DOCKERHUB_REGISTRY/$DOCKER_REPOSITORY:latest,$DOCKERHUB_REGISTRY/$DOCKER_REPOSITORY:$IMAGE_TAG
+
```
mediuminvader, line 447: Execute thin only when ubuntu x86 release is running

mediuminvader (changes requested):

mediuminvader (approved):

abhishekmishragithub:
This has been tested across `release-dev` branch to trigger the workflow.
- Here is the workflow log: https://github.com/getdozer/dozer/actions/runs/4945443681/jobs/8842178444
- You can cross-check the `containerimage.digest` value with pushes images on Docker hub 
- Screenshot from Docker hub

image

chubei:
macos workflow failed?

abhishekmishragithub:
> macos workflow failed?

Yeah, but I am unable to debug why, also we needed to do docker push only when `if: matrix.os == 'ubuntu-20-16-cores'`
it should skip for macos if that's the case.

chubei:
Included in #1558

PR #1555, opened 2023-05-11
Error Threshold Management

chloeminkyung:
when run: `dozer -- -e 10`

```bash
 INFO Starting gRPC server on http://0.0.0.0:50051 with security: None    
 INFO starting 8 workers
 INFO Tokio runtime found; starting in existing Tokio runtime
 INFO [pg] Starting replication
ERROR Sql: SQL Error: Division operation cannot be done.    
▹▹▹▹▹ err_threshold_test: 0: 0/s
▸▹▹▹▹ res: 1: 27.9485/s
▹▹▹▹▹ err_threshold_test: 0: 0/s
▹▹▹▹▹ res: 12: 330.1196/s
▹▹▹▹▹ err_threshold_test: 0: 0/s
```

based on: 
- https://github.com/getdozer/dozer/discussions/620
- https://github.com/getdozer/dozer/discussions/1237

Todos
- [x] adding err_threshold count in command option
- [x] adding err_threshold count logic in orchestrator
- [x] adding err type handling


dozer-cli/src/cli/types.rs
```diff
@@ -77,6 +79,13 @@ pub struct Deploy {
 pub struct App {
     #[command(subcommand)]
     pub command: AppCommands,
+    // #[arg(
+    //     global = true,
+    //     short = 'e',
+    //     long,
+    //     default_value = "0"
+    // )]
+    // pub err_threshold: String,
```
chubei, line 88: Remove unused comment?

dozer-cli/src/cli/types.rs
```diff
@@ -22,6 +22,8 @@ pub struct Cli {
     pub config_path: String,
     #[arg(global = true, long)]
     pub config_token: Option<String>,
+    #[arg(global = true, short = 'e', long, default_value = "0")]
+    pub err_threshold: String,
```
chubei, line 26: Can be a `u32` or something?

chubei (changes requested):
We should probably remove the `Option` error threshold and default to threshold 0

dozer-cli/src/cli/types.rs
chloeminkyung: oops sure

dozer-cli/src/cli/types.rs
chloeminkyung: just in case of people writing wrong values, it is better to parse it as String and setting default as 0 if it's invalid

chubei: I think it's better if people get noticed about their typo instead of we silently ignore that error and default to 0, and leave people wondering why their `21O` threshold didn't work.

chloeminkyung: what do you mean by 210? we set the max err threshold as well then?

chubei: Haha, that's the letter O followed by 21, used to explain the point.

chloeminkyung: Ah got it haha

dozer-cli/src/utils.rs
```diff
@@ -53,12 +53,12 @@ pub fn get_api_security_config(config: Config) -> Option<ApiSecurity> {
...
 }
 
-pub fn get_executor_options(config: &Config) -> ExecutorOptions {
+pub fn get_executor_options(config: &Config, err_threshold: Option<u32>) -> ExecutorOptions {
     ExecutorOptions {
         commit_sz: get_commit_size(config),
         channel_buffer_sz: get_buffer_size(config) as usize,
         commit_time_threshold: get_commit_time_threshold(config),
-        error_threshold: Some(0), // TODO: Read from configuration
+        error_threshold: err_threshold,
```
chubei, line 61: If the user doesn't say anything about error threshold, let's default to `0` to keep the original behaviour.

chubei (changes requested):

dozer-cli/src/utils.rs
chloeminkyung: alright fixed

chubei (approved):

PR #1557, opened 2023-05-12
feat: add cli dozer cloud login

duonganhthu43:


chubei:
Implemented in https://github.com/getdozer/dozer/pull/1604

PR #1608, opened 2023-05-29
feat: Check for running environment

mediuminvader:


chubei (commented):
I don't get this. Where's `DOZER_ENV` env var set?

mediuminvader:
> I don't get this. Where's `DOZER_ENV` env var set?

@chubei I'm assuming it's special var we're going to set up on the cluster, and not on local run.
Hence running on our cluster it will give the right value, running on local it just give back "local".

chubei:
How are we going to distinguish between if it's us developing or an actual user using it?

mediuminvader:
We don't need to recognize our development, just the cloud run.
But we could do that using a special VAR value to identify our own run.

chubei (approved):
Got it! Thanks for the explanation!

PR #1611, opened 2023-05-29
feat: emit Product processor metrics

mediuminvader:
Emit Join metrics on Prometheus endpoint

```

# HELP product_unsatisfied_joins Operations not matching the Join condition
# TYPE product_unsatisfied_joins counter
product_unsatisfied_joins 265

# TYPE product_output_operations counter
product_output_operations 1887841

# TYPE product_input_operations counter
product_input_operations 1888106

# HELP product_right_lookup_size Total number of items in the right lookup table
# TYPE product_right_lookup_size gauge
product_right_lookup_size 265

# HELP product_left_lookup_size Total number of items in the left lookup table
# TYPE product_left_lookup_size gauge
product_left_lookup_size 254

# HELP product_latency Processing latency
# TYPE product_latency summary
product_latency{quantile="0"} 0.0000025
product_latency{quantile="0.5"} 0.000040480100968151095
product_latency{quantile="0.9"} 0.000040480100968151095
product_latency{quantile="0.95"} 0.000040480100968151095
product_latency{quantile="0.99"} 0.000040480100968151095
product_latency{quantile="0.999"} 0.00014963672503375144
product_latency{quantile="1"} 0.028400292
product_latency_sum 17.38265707894686
product_latency_count 1888365

```

### DIFF
--- a/dozer-sql/src/pipeline/product/join/processor.rs
+++ b/dozer-sql/src/pipeline/product/join/processor.rs
@@ -18,30 +18,37 @@ pub struct ProductProcessor {
     join_operator: JoinOperator,
 }
 
+const LEFT_LOOKUP_SIZE: &str = "product.left_lookup_size";
+const RIGHT_LOOKUP_SIZE: &str = "product.right_lookup_size";
+const UNSATISFIED_JOINS: &str = "product.unsatisfied_joins";
+const IN_OPS: &str = "product.in_ops";
+const OUT_OPS: &str = "product.out_ops";
+const LATENCY: &str = "product.latency";
+
 impl ProductProcessor {
     pub fn new(join_operator: JoinOperator) -> Self {
         describe_gauge!(
-            "product.left_lookup_size",
+            LEFT_LOOKUP_SIZE,
             "Total number of items in the left lookup table"
         );
         describe_gauge!(
-            "product.right_lookup_size",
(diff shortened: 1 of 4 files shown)


dozer-sql/src/pipeline/product/join/processor.rs
```diff
@@ -16,6 +17,28 @@ pub struct ProductProcessor {
...
+        describe_gauge!(
+            "product.left_lookup_size",
+            "Total number of items in the left lookup table"
+        );
+        describe_gauge!(
+            "product.right_lookup_size",
+            "Total number of items in the right lookup table"
+        );
+        describe_gauge!(
+            "product.unsatisfied_joins",
```
chubei, line 29: Looks like this can only increase? In that case we can use a counter

chubei (changes requested):

dozer-sql/src/pipeline/product/join/processor.rs
mediuminvader: lookup table can decrease in case of TTL eviction

chubei: I'm talking about `unsatisfied_joins`

chubei (approved):
Do you think we need labels to distinguish between multiple joins?

PR #1619, opened 2023-05-31
chore: Inlcude `PodStatus` in `GetStatusResponse`

chubei:


chubei:
We're exposing these in the metrics API

PR #1629, opened 2023-06-07
chore: Add instance type to cloud config

chubei:


chubei:
@karolisg I don't think we need to address the e2e test error. Looks irrelavant.

karolisg (approved):

PR #1630, opened 2023-06-07
feature: Implement progress printer and disable time display

karolisg:


dozer-cli/src/console_helper.rs
```diff
@@ -1,8 +1,17 @@
-pub fn get_colored_text(value: &str, code: &str) -> String {
-    let mut result = String::from("\u{1b}[");
-    result.push_str(code);
-    result.push_str(";1m"); // End token.
-    result.push_str(value);
-    result.push_str("\u{1b}[0m");
-    result
+#[macro_export]
```
chubei, line 1: Let's keep it a function

chubei (changes requested):

dozer-cli/src/console_helper.rs
karolisg: Sure. Reverted change.

Issue #1653, opened 2023-06-19
WASM UDF support


snork-alt:
## 📝 Description:
This is a proposal for a new feature in Dozer Core, to enhance the functionality by allowing users to define a SQL User Defined Function (UDF) implemented in WASM (AssemblyScript)

## 🌟 Feature Details:
Users should be able to define a WASM UDF usable in Dozer SQL. 

In order to implement an UDF a user should:

- Add a `.wasm` file to the project folder (for example in the directory `udfs/wasm/my_function.wasm`)
- Add the following configuration lines to `dozer-config.yaml`:
```
udfs:
  wasm:
     my_function: udfs/wasm/my_function
```

Upon startup Dozer should load the wasm function and make it available in Dozer SQL. `my_function()` should now be usable from SQL as:

```
SELECT my_function(filed1, field2) FROM ...
```

hi-rustin:
I will give it a try.

tachyonicbytes:
Hi! I also wanted to explore this issue

/attempt #1653

hi-rustin:
@tachyonicbytes  Perhaps before contributing to the open source community one needs to ask the original people if they have started work. Otherwise, it's going to be an endless race for speed :(

varshith257:
/attempt #1653

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/varshith257) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @varshith257 | 1 bounty from 1 project | <div align="center">TypeScript, Go</div> | <div align="center">getdozer/dozer#1659</div> | [Cancel attempt](https://console.algora.io/api/bounties/clj2yqycv0002jz0f4dp6pr94/cancel-attempt) |

jamilahmadzai:
/attempt #1653

Plan:
- Add repo-native `!Wasm` UDF config support and schema coverage.
- Wire WASM UDF parsing through the SQL expression builder behind an optional `wasm` feature.
- Validate exported function signatures before execution and infer return/nullability metadata.
- Execute numeric WASM UDFs with checked casts, null propagation, and focused unit coverage for success/error paths.
- Propagate the optional feature through `dozer-sql`/`dozer-cli` so users can enable it cleanly.

ksushant6566:
/attempt #1653

[...]

Issue #1632, opened 2023-06-07
Feature Request: Inference with ONNX Model as Custom SQL Function


abhishekmishragithub:
## 📝 Description:

This is a proposal for a new feature in Dozer Core, to enhance the functionality by allowing users to utilize an ONNX model as a custom SQL function for scoring data in a streaming fashion. This feature would significantly benefit users who require real-time inference capabilities within their data processing pipelines.

## 🌟 Feature Details:

- **ONNX Model Integration:** Integrate the ONNX runtime into Dozer to support loading and executing ONNX models. This integration will enable users to leverage a wide range of pre-trained models from various popular machine learning frameworks.

- **Custom SQL Function:** Create a mechanism to define and register a custom SQL function using the configuration file. This function will serve as a bridge between the streaming data and the loaded ONNX model, allowing users to perform inference on the incoming data directly within their SQL queries.

- **Scalability and Performance:** Design the feature with scalability and performance in mind. Optimize the data processing pipeline to handle large volumes of streaming data efficiently, ensuring minimal latency and high throughput.

- **Error Handling and Fault Tolerance:** Leverage on Dozer error handling mechanisms to gracefully handle exceptions during the inference process.

- **Documentation and Examples:** Provide comprehensive documentation and examples to guide users on how to utilize this new feature effectively. The documentation should cover the integration steps, usage guidelines, and best practices for implementing and deploying the streaming inference pipeline.

## ✅ Benefits:

Users can leverage the power of ONNX models to perform real-time inference within their SQL queries. Users can seamlessly incorporate machine learning capabilities into their existing Dozer runtime without the need for extensive code changes or complex integrations. The feature allows users to select and use a wide range of pre-trained models.

ibrahim-akrab:
I would like to take a stab at this, however, there is barely any documentation for dozer-core that I could find.

chubei:
This boils down to add a variant to [`Expression`](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/expression/execution.rs#L25) in `dozer_sql`,  which can be further divided into 3 tasks.

[...]

PR #1654, opened 2023-06-19
Add incipient wasm_udf support

tachyonicbytes:
This PR adds incipient support for wasm_udf, as defined in TODO: issue. A new feature is added, so compile with `--features wasm`, if you want to test it.

I used Wasmtime as the wasm engine to run the wasm code. Wasmer or WasmEdge or some other engine can be chosen as well, with not many modifications, in theory.

The most important part is the type conversion from Dozer types to wasm types, which natively are very few (only i64 and f64 could be used for now). Wasm doesn't natively support i128, and Dozer does not support i32, so there were not many type that matched.

More types can of course be added (Boolean for example is trivial to add, even if it wastes an entire i32). For composite types, like String or Text, binding generation support has to be added.

I could not find the parsing of the `dozer-config.yml`, so I was inspired by the python_udf module to use environment variables. You can define `DOZER_WASM_UDF` to point to a `.wasm` file with the exported functions.

Testplan:

An AssemblyScript example has been added to `./dozer-tests/wasm_udf/assemblyscript`. The simples way to test the feature is to use the `dozer-samples` sql join test.

Change the sql to include a simple wasm_udf call:

```
sql: |
  SELECT t.tpep_pickup_datetime AS pickup_time, z.Zone AS zone, wasm_addf(9.5, 10.7, 'float') as WASM
  INTO pickup
  FROM trips t JOIN taxi_zone_lookup z ON t.PULocationID = z.LocationID;
```
Follow the rest of the instructions in that example.

To build the AssemblyScript that contains the `wasm_addf` function:

```
cd ./dozer-tests/wasm_udf/assemblyscript
npm install && npm run asbuild
```

The module is now at `./dozer-tests/wasm_udf/assemblyscript/build/debug.wasm`. Set `DOZER_WASM_UDF` before you run the example:

```
DOZER_WASM_UDF=./path/to/debug.wasm dozer
```

/claim #1653


snork-alt:
Thanks. Can you review @chubei ?

chubei:
> Thanks. Can you review @chubei ?

Sure.

snork-alt:
@tachyonicbytes what do you mean you could not find the parsing of `dozer-config.yml`? @chubei can guide you here. We are trying to centralize the UDFs definitions. Check out also #1632 as well for UDFs definition

tachyonicbytes:
@snork-alt I checked #1632 while developing. Is the definition you refer to the
```
udfs:
  wasm:
    my_function: module.wasm
```
part of the config?

In that case, that is precisely what I didn't get. It's not the parsing itself that gets me @chubei, but I don't know what I should import to get the module path `"path/to/debug.wasm"` in `wasm_udf.rs`. The python module didn't have it, it used env vars, and the `onnx` module is not ready yet, so I couldn't see how they do it.

snork-alt:
@chubei, can you help @tachyonicbytes? Thanks

chubei (commented):
Hi @tachyonicbytes, this is great work and your PR summary is very helpful. I see two places where we can improve in this PR. Please let me know what you think.

1. Instead of using environment variables and `wasm_`prefix, we should use a designated config file section to define the UDFs, as @snork-alt suggested.

I understand that you're taking the Python UDF implementation as a reference. The truth is we've learnt more since we merged the Python UDF and we have a clearer understanding of the UDF usage pattern now so we decided to put the UDF definitions in the config file.

You mentioned the problem is that, you don't know how you can get access to the wasm module path defined in the config file. I believe we can add a parameter `udfs` to [`statement_to_pipeline`](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/builder.rs#L63), passing it all the way down to [`parse_sql_function`](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/expression/builder.rs#L384), and there we can look up the function name to determine if it's a UDF, and if it is, which kind of UDF.

2. We need to implement double direction data type conversion.

As you mentioned, natively few types are matched. The goal here is to support as many types as possible.

For example, if a wasm UDF returns `i32`, we should convert it to `i64` to use it in Dozer. Here the principle is no data loss, so we won't convert Dozer `i128` to wasm, which we just don't support.

We should also write `string`, `timestamp`, etc. conversions, so authors of the UDF gets a easy to use type.

I suggest we think about data type conversion thoroughly and produce a design document outlining what we'll do. Once we agree on the design we can get into implementation.

As a reference, you can check https://github.com/getdozer/dozer/pull/1200 to see how we handle `json` type conversion between Dozer and Python, Dozer and Arrow, etc.

That's all my comments! Thank you for your great work again!

snork-alt:
@tachyonicbytes are you planning to implement those changes and close the PR ?

tachyonicbytes:
Thank you so much for the review, @chubei and @snork-alt.

Ok, so I understand that, in order to pass the config variables to the wasm_udf module, we need to propagate those values through the necessary functions. I thought that the project had something akin to a global config module that I import and use. That's what I was looking for, and could not find.

Yeah, the python_udf module has been a semi-reference for me, because when I looked at it in more depth, it seemed incomplete. But it was still a nice starting point.

Now, for the data type conversion:

So, for WebAssembly, this is what we work with: i32, i64, f32, f64, v128, and linear memory for arbitrary amounts of memory. In Dozer, we have (please correct me if I am wrong with any of it):

- [ ] UInt
- [ ] U128
- [x] Int
- [ ] I128
- [x] Float
- [ ] Boolean
- [ ] String
- [ ] Text
- [ ] Binary
- [ ] Decimal
- [ ] Timestamp
- [ ] Date
- [ ] Json

Int and Float are done. Boolean is trivial, albeit it wastes 31 bits for every type. UInt should be doable, even though we may have to impose certain restrictions on its usage. String, Text, Binary all involve the liniar memory in some capacity, so you kinda have to solve them all at once, so it's only one data type's worth of work, plus idiosyncrasies.

Now, there is a catch. In theory, U128 and I128 are easily solvable with v128, which, even though is a simd type, can support the usual scalar operations that we do on I128 and U128 (with the exception of [division](https://stackoverflow.com/a/75495728/20610519), which can be done slower or it can be restricted). But the [wasmtime](https://docs.rs/wasmtime/latest/wasmtime/trait.WasmTy.html#) crate does not implement v128. It is in AssemblyScript, but we can't use it because of that. The [wasmer](https://docs.rs/wasmer/latest/wasmer/enum.Type.html) crate, on the other hand, supports it.

If this is solved, UInt can be half of a v128, even though that is, again, not fully efficient and I128 and U128 have the caveat above, no or slow division.

I don't know how Decimal works inside Dozer. Is it an Int type that you don't fully use? That's how it was implemented in intro CS classes, if I recall correctly. If so, that is an easy conversion, but it's not natively supported in any langauge, so users will have to implement their own semantics. If it is implemented as a String, then it falls in the linear memory category.

Timestamp and Date should be strings, I guess, so linear memory for them as well.

The last one, Json. This is the one that I would personally suggest skipping at least for now. The problem with WebAssembly and json is not that people don't need it, but it is kinda hard to do without proper reflection. AssemblyScript has no native json support. It has a json package that is great in its own way, but it doesn't support exception handling, so it's kinda hard to validate your json on the udf end. TinyGo does not support the standard library json, although it has some non-reflective packages. 

Of course, there is also the possibility of splitting the data type conversion by parameter or return value. We can decide to implement Json for parameters, but not for return values, for example, so you always have valid jsons in your udf code.

So, please let me know which of the types above sound doable for this PR. Please let me know of the wasmtime <-> wasmer decision. And please further explain what is needed of Timestamp and Date.

I'll also repair the python_udf module and add a TinyGo wasm example, if you don't mind.

dozer-sql/src/pipeline/expression/builder.rs
```diff
@@ -396,6 +396,13 @@ impl ExpressionBuilder {
             return self.parse_python_udf(udf_name, sql_function, schema);
         }
 
+        #[cfg(feature = "wasm")]
```
snork-alt, line 399: Would be nice if instead of having a `wasm_` prefix, wasm based UDFs are declared in the config file. Do not follow  too much the Python sample as we have to refactor it.

snork-alt (changes requested):

snork-alt:
@tachyonicbytes we just merged a PR that allows return types of SQL UDFs to be specified as `function<Type>`. This can be useful for WASM functions.

tachyonicbytes:
Perfect. Less matching on strings for `wasm_udfs` as well

snork-alt:
@tachyonicbytes are you still working on this ?

tachyonicbytes:
Yes. So the config part seems to be more complicated than I anticipated. The problem is that a new section has to be created for wasm udfs. I added it to `api_config`, because it seems to have more in common with the other `api_config` options than the others.

Another problem is that, in the `ExecutorOptions` at `dozer-core/src/executor.rs` the wasm udfs need to be added. I opted for this: `pub udfs: HashMap<String, HashMap<String, String>>`. But the `HashMap` is not serializable with `prost`, so another solution has to be found, as all the other types are simple types.

Changing that type also means going back to the `UdfOptions` struct I put in `dozer-types/src/models/api_config.rs` and change the type there as well.

The problem is that the free format

```
wasm_udfs:
    - module1: function1
    - module1: function2
    - module2: function3
```

is hard to serialize to simple types, if you don't store it in a string, for example, in an ad-hoc format, like `"module1|function1|module1|function2"` etc. But maybe that's an option as well.

Is there a discord server or some synchronous communication mechanism for outsiders that contribute to dozer? I think some of the problems that I face can be easily solved that way.

chubei:
Hi @tachyonicbytes , let's talk on Discord. https://discord.gg/3eWXBgJaEQ

And we've added `udfs` config section in https://github.com/getdozer/dozer/pull/1831. You can follow format in that PR.

snork-alt:
@tachyonicbytes any plan for.this ?

snork-alt:
Hi @tachyonicbytes is this ready for review ? If yes please ask @chubei . Thanks

tachyonicbytes:
@snork-alt I talked with @chubei, and the agreed strategy is to wait for this https://github.com/getdozer/dozer/pull/1838 to be merged and then rebase wasm udfs upon it. Otherwise it would contain a lot of duplicated work that would need to be resolved in conflicts.

.github/workflows/coverage.yaml
```diff
@@ -121,7 +121,9 @@ jobs:
         shell: bash
         run: |
           source ./dozer-tests/python_udf/virtualenv.sh
-          cargo test --features snowflake,ethereum,kafka,python,mongodb --no-fail-fast
+          apt-get update && apt-get install -y npm
+          (cd ./dozer-tests/wasm_udf/assemblyscript && npm install && npm run asbuild)
```
tachyonicbytes, line 125: This is built here, but not used yet. It's here to be used as a resource for tests that want to run wasm_udfs.

tachyonicbytes:
The PR is functional now. I usually check it with the `dozer-samples/sql/join` sample:

```
version: 1
app_name: sql-join-sample
connections:
  - config : !LocalStorage
      details:
        path: data
      tables:
        - !Table
          name: taxi_zone_lookup
          config: !CSV
            path: zones
            extension: .csv
        - !Table
          name: trips
          config: !Parquet
            path: trips
            extension: .parquet
    name: ny_taxi

sql: |
  SELECT t.tpep_pickup_datetime AS pickup_time, z.Zone AS zone, fib<Int>(10) as WASM
  INTO pickup
  FROM trips t JOIN taxi_zone_lookup z ON t.PULocationID = z.LocationID;

sources:
  - name: taxi_zone_lookup
    table_name: taxi_zone_lookup
    connection: ny_taxi
  - name: trips
    table_name: trips
    connection: ny_taxi

endpoints:
  - name: pickup
    path: /pickup
    table_name: pickup

udfs:
  - config: !Wasm
      path: /path/to/dozer/dozer-tests/wasm_udf/assemblyscript/build/debug.wasm
    name: fib
```

snork-alt:
@tachyonicbytes is it possible to infer return type and parameter types from the wasm module itself, rather than having to declare it during the function call ?

tachyonicbytes:
I am 90% sure that you can infer it, but I will have to thoroughly check the docs for that. The other thing is that `return_type` is necessary there so that Dozer requests from the wasm module the actual type it wants, so that there is less confusion during the conversion.

snork-alt:
@chubei can you review ?

chubei:
> @chubei can you review ?

I'm leaving it for later of this week.

dozer-sql/src/pipeline/errors.rs
```diff
@@ -13,6 +13,8 @@ use std::fmt::{Display, Formatter};
 
 #[cfg(feature = "onnx")]
 use crate::pipeline::onnx::OnnxError;
+#[cfg(feature = "wasm")]
+use wasmtime;
```
chubei, line 17: Is this `use` used?

dozer-sql/src/pipeline/expression/wasm_udf.rs
```diff
@@ -0,0 +1,76 @@
...
+    args: &[Expression],
+    return_type: &FieldType,
+    record: &Record,
+) -> Result<Field, PipelineError> {
+    let values = args
+        .iter()
+        .map(|arg| arg.evaluate(record, schema))
+        .collect::<Result<Vec<_>, PipelineError>>()?;
+
+    let values2: Vec<Val> = values
```
chubei, line 24: You can just use `let values` to shadow the previous variable.

dozer-sql/src/pipeline/expression/wasm_udf.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let engine = Engine::default();
+
+    let module = Module::from_file(&engine, module)?;
+    let mut store = Store::new(&engine, 4);
+    let instance = Instance::new(&mut store, &module, &[])?;
+
+    let wasm_udf_func = instance
+        .get_func(&mut store, name)
+        .expect("export wasn't a function");
```
chubei, line 41: Initialization code should go to expression building?

dozer-sql/src/pipeline/expression/wasm_udf.rs
```diff
@@ -0,0 +1,76 @@
...
+            Field::Int(value) => Val::I64(*value),
+            Field::Float(value) => Val::F64(value.to_bits()),
+            _ => todo!(),
+        })
+        .collect();
+
+    let engine = Engine::default();
+
+    let module = Module::from_file(&engine, module)?;
+    let mut store = Store::new(&engine, 4);
```
chubei, line 36: What does the magic number `4` mean?

dozer-sql/src/pipeline/expression/wasm_udf.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::pipeline::errors::PipelineError::UnsupportedSqlError;
+use crate::pipeline::errors::UnsupportedSqlError::GenericError;
+use crate::pipeline::expression::execution::Expression;
+
+use dozer_types::ordered_float::OrderedFloat;
+use dozer_types::types::{Field, FieldType, Record, Schema};
+
+use wasmtime::*;
+
+pub fn evaluate_wasm_udf(
```
chubei, line 11: We need to remove any `todo!` or `expect` from this function

dozer-types/src/tests/udf_yaml_deserialize.rs
```diff
@@ -17,3 +17,22 @@ fn standard() {
     let expected = udf_conf;
     assert_eq!(expected, deserializer_result);
 }
+
+#[cfg(feature = "wasm")]
+#[test]
+fn standard_wasm() {
```
chubei, line 23: This test doesn't seem to pass

chubei (commented):
This PR looks good in general. We need to improve a little in error handling but that should be trivial.

I think there're still two things missing now: validation and casting.

# Validation

As you know, wasm function is [statically typed](https://docs.rs/wasmtime/latest/wasmtime/struct.Func.html#method.ty), and every Rust type we support right now can be mapped uniquely to a wasm type. So it's necessary to validate:

- Does the input schema matches wasm function params?
- Does the wasm function result have exactly one output?

And we should infer return type instead of asking user to write it in the function signature.

# Casting

To support as many types as possible, casting should be performed. If dozer type is `U128` or `I128`, we should cast it to wasm `i64`. It should emit a precision loss warning during validation, and a runtime error if the cast cannot be performed due to overflow at runtime.

Let's skip the support for non-primitive types in this PR.

Onnx module has all these validation and casting functions. You can take that as a reference.

snork-alt:
@tachyonicbytes can we fix the last things so we can merge ? Thanks

snork-alt:
@tachyonicbytes shall we close this or you plan to wrap it up so it can be merged and your bounty paid?

tachyonicbytes:
Yeah, sorry for the delay, I'm wrapping it up.

snork-alt:
Any progress @tachyonicbytes ?

tachyonicbytes:
Yes, I've updated to the new json schemas, as well as working on the new errors, like `onnx` does. With testing and all it should be ready in maybe a couple of days, maybe Sunday.

dozer-sql/expression/src/builder.rs
```diff
@@ -489,6 +492,26 @@ impl ExpressionBuilder {
...
+                    }
+
+                    #[cfg(not(feature = "wasm"))]
+                    {
+                        let _ = config;
+                        Err(Error::WasmNotEnabled)
+                    }
+
+                }
+                _ => Err(Error::UnsupportedUdfType),
```
chubei, line 514: We shouldn't need this line?

dozer-sql/src/errors.rs
```diff
@@ -54,6 +54,13 @@ pub enum PipelineError {
     #[error("Duplicate INTO table name found: {0:?}")]
     DuplicateIntoClause(String),
 
+    // #[cfg(feature = "wasm")]
+    // #[error("Wasm Error: {0}")]
+    // WasmErr(wasmtime::Error),
+    // #[cfg(not(feature = "wasm"))]
+    // #[error("Wasm UDF feature is not enabled")]
+    // WasmNotEnabled,
+
```
chubei, line 63: Let's remove these lines?

dozer-sql/src/errors.rs
```diff
@@ -131,6 +138,13 @@ pub enum PipelineError {
     ProcessorAlreadyExists(String),
 }
 
+// #[cfg(feature = "wasm")]
+// impl From<wasmtime::Error> for PipelineError {
+//     fn from(wasm_err: wasmtime::Error) -> Self {
+//         PipelineError::WasmErr(wasm_err)
+//     }
+// }
+
```
chubei, line 147: Let's remove these lines?

dozer-sql/src/pipeline/expression/mod.rs
chubei, line 1: Is this an unused file?

dozer-sql/Cargo.toml
```diff
@@ -29,3 +30,4 @@ tokio = { version = "1", features = ["rt", "macros"] }
 [features]
 python = ["dozer-sql-expression/python"]
 onnx = ["dozer-sql-expression/onnx"]
+wasm = ["wasmtime"]
```
chubei, line 33: Looks like we should remove `wasmtime` dependency and change it to `dozer-sql-expression/wasm`?

chubei (commented):
This looks great! I left some comments in the code, but they are all minor issues.

Questions:

- Looks like we're missing a `wasm/utils.rs` file?
- Let's change the `todo!`s to `Err`s.
- Let's also validate the input argument conversion before execution.

dozer-sql/Cargo.toml
```diff
@@ -19,6 +19,7 @@ enum_dispatch = "0.3.12"
 linked-hash-map = { version = "0.5.6", features = ["serde_impl"] }
 metrics = "0.21.0"
 multimap = "0.9.0"
+wasmtime = { version = "9.0.4", optional = true }
```
chubei, line 22: We don't need this dependency?

dozer-sql/expression/src/wasm/udf.rs
```diff
@@ -0,0 +1,98 @@
...
+        Some(func) => {
+            wasm_udf_func = func;
+        }
+        None => {
+            return Err(Wasm(WasmFunctionMissing(
+                name.to_string(),
+                config.to_string(),
+            )));
+        }
+    }
```
chubei, line 38: These code should go into the compilation phase? Just like the onnx UDF is passing in a `Session` instead of the model path.

dozer-sql/expression/src/wasm/udf.rs
```diff
@@ -0,0 +1,98 @@
...
+    args: &[Expression],
+    record: &Record,
+) -> Result<Field, Error> {
+    let input_values = args
+        .iter()
+        .map(|arg| arg.evaluate(record, schema))
+        .collect::<Result<Vec<_>, Error>>()?;
+
+    let engine = Engine::default();
+    let module = Module::from_file(&engine, config).unwrap();
```
chubei, line 23: Cannot `unwrap`

dozer-sql/expression/src/wasm/udf.rs
```diff
@@ -0,0 +1,98 @@
...
+    let mut store = Store::new(&engine, ());
+    let instance = Instance::new(&mut store, &module, &[]).unwrap();
```
chubei, line 25: Cannot `unwrap`

chubei (changes requested):
Looks like `wasm/utils.rs` is missing?

We'll need at least one test to show that the UDF actually runs.

Besides, is it possible to use `deno` instead of `wasmtime`? As we are already depending on `deno`.

dozer-sql/Cargo.toml
tachyonicbytes: This is the wasm runtime that we use. This runs the actual wasm functions

PR #1658, opened 2023-06-20
chore: add name fn for ProcessorFactory for UIDag in Cloud

duonganhthu43:


dozer-core/src/node.rs
```diff
@@ -70,6 +70,7 @@ pub trait ProcessorFactory<T>: Send + Sync + Debug {
         input_schemas: HashMap<PortHandle, Schema>,
         output_schemas: HashMap<PortHandle, Schema>,
     ) -> Result<Box<dyn Processor>, BoxedError>;
+    fn name(&self) -> String;
```
chubei, line 73: Let's rename to `typ` or `type_`. `name` often refers to the node's name (`NodeHandle`) in our context.

chubei (changes requested):

chubei (approved):

PR #1672, opened 2023-06-23
feat: pg tls config connectivity

chloeminkyung:
`sslmode: disable | Disable`
`sslmode: prefer | Prefer`
`sslmode: require | Require`

disable = tokio-postgres::Config::SslMode::Disable + rustls no verification
prefer = tokio-postgres::Config::SslMode::Prefer + rustls no verification
verify-full = tokio-postgres::Config::SslMode::Require + rustls verification

disable - no validation at all, no need to use TLS
Prefer - no validation at all, try using TLS, but if it fails, back to disable mode
verify-full - validation on connection to the server using encryption, only connects to the server that I specify and it should be the one that I already trust, if fails, fail the whole process


dozer-cli/src/cli/init.rs
```diff
@@ -106,11 +106,13 @@ pub fn generate_connection(connection_name: &str) -> Connection {
...
-                host: "localhost".to_owned(),
-                port: 5432,
-                database: "users".to_owned(),
+                user: Some("postgres".to_owned()),
+                password: Some("postgres".to_owned()),
+                host: Some("localhost".to_owned()),
+                port: Some(5432),
+                database: Some("users".to_owned()),
+                sslmode: None,
+                connection_url: Some("".to_owned()),
```
chubei, line 115: Should this be `None` to enable those `user`, `password` options?

dozer-ingestion/src/errors.rs
```diff
@@ -123,6 +123,9 @@ pub enum ConfigurationError {
 
 #[derive(Error, Debug)]
 pub enum PostgresConnectorError {
+    #[error("Invalid SslMode: {0}")]
+    InvalidSslError(String),
```
chubei, line 127: Let's put `SslMode` inside instead of a `String`.

dozer-tests/src/e2e_tests/runner/running_env.rs
```diff
@@ -338,8 +338,8 @@ fn write_dozer_config_for_running_in_docker_compose(
 
         match config {
             ConnectionConfig::Postgres(postgres) => {
-                postgres.host = connection.name.clone();
-                postgres.port = map_port(postgres.port as u16) as u32;
+                postgres.host = Some(connection.name.clone());
```
chubei, line 341: Should we `replenish` this config?

dozer-types/src/models/app_config.rs
```diff
@@ -52,7 +52,7 @@ pub struct Config {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     #[prost(message, tag = "10")]
-    /// flags to enable/disable features
+    /// flags to enable/Disable features
```
chubei, line 55: Revert this?

dozer-types/src/models/connection.rs
```diff
@@ -16,30 +20,175 @@ pub struct Connection {
...
             ["port", self.port],
-            ["database", self.database]
+            ["database", self.database],
+            ["sslmode", format!("{:?}", self.ssl_mode)]
         )
     }
 }
 
+impl PostgresConfig {
+    pub fn replenish(&self) -> PostgresConfigReplenished {
```
chubei, line 63: Should we return an error instead of `unwrap`? With helpful error messages like `user` must be present if `connection_url` is not present.

dozer-types/src/models/connection.rs
```diff
@@ -16,30 +20,175 @@ pub struct Connection {
...
+    }
+}
+
+fn get_ssl_mode(mode: Option<String>) -> SslMode {
+    match mode {
+        Some(m) => match m.as_str() {
+            "disable" | "Disable" | "" => SslMode::Disable,
+            "prefer" | "Prefer" => SslMode::Prefer,
+            "require" | "Require" => SslMode::Require,
+            &_ => SslMode::Disable,
```
chubei, line 109: I don't like the idea of treating everything as `disable`. For example if user makes a typo `perfer` he'll be wondering why it doesn't work.

I think the better way is to return an error saying `Unknown SSL mode {m}`

dozer-types/src/models/connection.rs
```diff
@@ -16,30 +20,175 @@ pub struct Connection {
...
+        },
+        None => SslMode::Disable,
+    }
+}
+
+pub fn connection_url_map(url: &str, config: &PostgresConfig) -> HashMap<String, String> {
```
chubei, line 115: I'm not able to read this regex. Is there a library for parsing the connection string?

dozer-types/src/tests/dozer_yaml_deserialize.rs
```diff
@@ -137,32 +137,3 @@ fn error_missing_field_connection_ref_in_source() {
         .to_string()
         .starts_with("sources[0]: missing connection ref"));
 }
-
-#[test]
-fn error_missing_field_inner() {
```
chubei, line 142: We should keep this test

chubei (changes requested):
Great job! The comments are only about style and error handling.

dozer-cli/src/cli/init.rs
chloeminkyung: right!

dozer-ingestion/src/errors.rs
chloeminkyung: Sure

dozer-tests/src/e2e_tests/runner/running_env.rs
chloeminkyung: right will do

dozer-types/src/models/app_config.rs
chloeminkyung: Oh didn't know I changed this part, will revert

dozer-types/src/models/connection.rs
chloeminkyung: Can we do `.expect()` with helpful error message instead?

chubei: That's kind of inconsistent with what we're doing now. Why we need to use `expect`?

dozer-types/src/models/connection.rs
chloeminkyung: Ah got it! sure will do the change

dozer-types/src/models/connection.rs
chubei: > Configuration can be parsed from libpq-style connection strings

https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html#impl-FromStr-for-Config

dozer-types/src/models/connection.rs
chloeminkyung: to get rid of extra layer of Result checking with .unwrap()

chubei: I can think of two downsides of getting rid of `Result`:

1. Inconsistency with error handling elsewhere.
2. Make it difficult if upper layer wants to handle the error originated in `replenish`.

What's the benefit here?

dozer-ingestion/src/errors.rs
chubei: We can do `#[error("Invalid SslMode: {0:?}")]`. That's what the current error message is anyway.

dozer-types/src/tests/dozer_yaml_deserialize.rs
```diff
@@ -156,13 +156,13 @@ fn error_missing_field_inner() {
...
+      - phone
   "#;
     let deserialize_result = serde_yaml::from_str::<Config>(input_config);
     let error = deserialize_result.err();
     assert!(error.is_some());
     assert!(error
         .unwrap()
         .to_string()
-        .starts_with("connections[0].config: missing field `password`"));
+        .starts_with("sources[0]: missing connection ref"));
```
chubei, line 167: I think this test is meant to check that, when `password` is missing, it's properly reported.

chubei (changes requested):

dozer-types/src/tests/dozer_yaml_deserialize.rs
chloeminkyung: The reason why we are getting that is due to, cause we change that to a optional variable, it will create the Config variable. It used to fail without creating them.

dozer-ingestion/src/errors.rs
chloeminkyung: Oh! didn't know we can do that haha will change it that way

dozer-types/src/tests/dozer_yaml_deserialize.rs
chubei: Got it. Then maybe change the test name? It's actually testing `missing connection ref` now.

dozer-types/src/models/connection.rs
chloeminkyung: converting the function to return with `Result`

dozer-types/src/tests/dozer_yaml_deserialize.rs
chloeminkyung: got it!

dozer-ingestion/src/connectors/mod.rs
```diff
@@ -231,7 +231,9 @@ pub fn get_connector(connection: Connection) -> Result<Box<dyn Connector>, Conne
 
 pub fn get_connector_info_table(connection: &Connection) -> Option<Table> {
     match &connection.config {
-        Some(ConnectionConfig::Postgres(config)) => Some(config.convert_to_table()),
+        Some(ConnectionConfig::Postgres(config)) => {
```
chubei, line 234: Propagate error instead of `unwrap`

dozer-ingestion/src/connectors/postgres/connection/helper.rs
```diff
@@ -1,25 +1,48 @@
...
+use std::sync::Arc;
+use std::time::SystemTime;
+use tokio_postgres::config::SslMode;
+use tokio_postgres::{Client, NoTls};
 
 pub fn map_connection_config(
     auth_details: &ConnectionConfig,
 ) -> Result<tokio_postgres::Config, ConnectorError> {
     if let ConnectionConfig::Postgres(postgres) = auth_details {
+        let config_replenished = postgres.replenish().unwrap();
```
chubei, line 16: Propagate error instead of `unwrap`

dozer-types/src/errors/types.rs
```diff
@@ -59,4 +59,12 @@ pub enum DeserializationError {
     Utf8(#[from] std::str::Utf8Error),
     #[error("Failed to convert type due to json numbers being out of the f64 range")]
     F64TypeConversionError,
+    #[error("Unknown SSL mode: {0}")]
+    UnknownSslMode(String),
+    #[error("Invalid connection url for Postgres configuration")]
+    InvalidConnectionUrl,
+    #[error("{0} is missing in Postgres configuration")]
+    MissingFieldInPostgresConfig(String),
+    #[error("{0} is mismatching in Postgres configuration")]
```
chubei, line 68: "mismatching" or "conflicting"?

dozer-types/src/errors/types.rs
```diff
@@ -59,4 +59,12 @@ pub enum DeserializationError {
     Utf8(#[from] std::str::Utf8Error),
     #[error("Failed to convert type due to json numbers being out of the f64 range")]
     F64TypeConversionError,
+    #[error("Unknown SSL mode: {0}")]
+    UnknownSslMode(String),
+    #[error("Invalid connection url for Postgres configuration")]
+    InvalidConnectionUrl,
```
chubei, line 65: We need to include the error source and tell the user why the connection url is invalid

dozer-types/src/models/connection.rs
```diff
@@ -16,30 +26,222 @@ pub struct Connection {
...
+fn get_sslmode(mode: String) -> Result<SslMode, DeserializationError> {
+    match mode.as_str() {
+        "disable" | "Disable" => Ok(SslMode::Disable),
+        "prefer" | "Prefer" => Ok(SslMode::Prefer),
+        "require" | "Require" => Ok(SslMode::Require),
+        &_ => Err(UnknownSslMode(mode)),
+    }
+}
+
+pub fn connection_url_map(url: &str, config: &PostgresConfig) -> HashMap<String, String> {
```
chubei, line 168: Looks like this function is unused?

chubei (commented):
Looking great! Only have questions about a few places.

dozer-ingestion/src/connectors/mod.rs
chloeminkyung: How are we propagate the error if the return type is not having Result? panic?

dozer-ingestion/src/connectors/postgres/connection/helper.rs
chloeminkyung: Sure

dozer-ingestion/src/connectors/mod.rs
chubei: Yes changing the return type to `Result`. That's what I wanted to say. Sorry didn't make it clear.

dozer-types/src/errors/types.rs
chloeminkyung: following the convention that we have been using the word "mismatching"

dozer-types/src/errors/types.rs
chloeminkyung: sure

dozer-types/src/models/connection.rs
chloeminkyung: forgot to remove! will remove

dozer-ingestion/src/connectors/mod.rs
```diff
@@ -229,15 +230,18 @@ pub fn get_connector(connection: Connection) -> Result<Box<dyn Connector>, Conne
...
-        Some(ConnectionConfig::Postgres(config)) => Some(config.convert_to_table()),
-        Some(ConnectionConfig::Ethereum(config)) => Some(config.convert_to_table()),
-        Some(ConnectionConfig::Snowflake(config)) => Some(config.convert_to_table()),
-        Some(ConnectionConfig::Kafka(config)) => Some(config.convert_to_table()),
-        Some(ConnectionConfig::S3Storage(config)) => Some(config.convert_to_table()),
-        Some(ConnectionConfig::LocalStorage(config)) => Some(config.convert_to_table()),
-        _ => None,
+        Some(ConnectionConfig::Postgres(config)) => match config.replenish() {
+            Ok(conf) => Ok(conf.convert_to_table()),
+            Err(_) => Err(WrongConnectionConfiguration),
```
chubei, line 237: As stated before, we should try to include error context and source in an error variant, so it's actually helpful.

dozer-ingestion/src/connectors/postgres/connection/helper.rs
```diff
@@ -1,25 +1,52 @@
...
+use tokio_postgres::config::SslMode;
+use tokio_postgres::{Client, NoTls};
 
 pub fn map_connection_config(
     auth_details: &ConnectionConfig,
 ) -> Result<tokio_postgres::Config, ConnectorError> {
     if let ConnectionConfig::Postgres(postgres) = auth_details {
+        let config_replenished = match postgres.replenish() {
+            Ok(conf) => conf,
+            Err(_) => return Err(WrongConnectionConfiguration),
```
chubei, line 19: Same as above

dozer-types/src/errors/types.rs
```diff
@@ -59,4 +59,12 @@ pub enum DeserializationError {
     Utf8(#[from] std::str::Utf8Error),
     #[error("Failed to convert type due to json numbers being out of the f64 range")]
     F64TypeConversionError,
+    #[error("Unknown SSL mode: {0}")]
+    UnknownSslMode(String),
+    #[error("Invalid connection url for Postgres configuration: {0}")]
+    InvalidConnectionUrl(String),
```
chubei, line 65: It's better if we can keep the original error enum/struct instead of converting it to string, so we don't lose information.

chubei (changes requested):

dozer-ingestion/src/connectors/mod.rs
chloeminkyung: sure

dozer-ingestion/src/connectors/postgres/connection/helper.rs
chloeminkyung: sure

dozer-types/src/errors/types.rs
chloeminkyung: Sure

chubei (approved):
Thank you!

PR #1674, opened 2023-06-23
feat: qualified wildcard

chloeminkyung:
```sql
  SELECT
    users.*
  INTO res 
  FROM users;
  
  SELECT 
    phones.* 
  INTO join_res
  FROM users
  INNER JOIN phones
    ON phones.call = users.phone
  WHERE users.age > 12;
```

### DIFF
--- a/dozer-tests/README.md
+++ b/dozer-tests/README.md
@@ -53,6 +53,8 @@ Otherwise, if `error.json` is found, the framework expects `dozer` fails to star
 
 For all supported error expectation checks, see `expectations.rs`.
 
+In case of addition of new test case, be aware of non-colliding ports from the connection that you are creating for each test cases.
+
 ## Add a Connection
 
 Test cases may need to establish various kinds of connections. Connection services can run as a docker compose service.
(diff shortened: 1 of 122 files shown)


dozer-tests/README.md
```diff
@@ -53,6 +53,8 @@ Otherwise, if `error.json` is found, the framework expects `dozer` fails to star
 
 For all supported error expectation checks, see `expectations.rs`.
 
+In case of addition of new test case, be aware of non-colliding ports from the connection that you are creating for each test cases.
```
chubei, line 56: This is not the correct place to document this issue. This README describes the end-to-end test framework, not the unit tests that start with `test_e2e_`

chubei (changes requested):

dozer-tests/README.md
chloeminkyung: Ah wrong place! will replace it

mediuminvader (approved):

chubei (approved):

PR #1684, opened 2023-06-27
chore: group CUD operation in cache/source as single metric 

duonganhthu43:


dozer-cli/src/pipeline/connector_source.rs
```diff
@@ -314,23 +313,23 @@ impl Source for ConnectorSource {
...
 
-                    let labels = [
+                    let mut labels = vec![
                         ("connection", self.connection_name.clone()),
                         ("table", table_name.to_string()),
                     ];
                     match &kind {
                         IngestionMessageKind::OperationEvent(Operation::Delete { .. }) => {
-                            increment_counter!(SOURCE_DELETE_COUNTER_NAME, &labels);
+                            labels.push(("operation_type", "delete".to_owned()))
```
chubei, line 322: We don't need these `to_owned`

chubei (approved):

dozer-cli/src/pipeline/connector_source.rs
duonganhthu43: since the labels have type `Vec<(&str, String)>` we have to convert it anyway

chubei: Sure my bad.

PR #1686, opened 2023-06-27
fix: avoid include snapshot event in metrics count

duonganhthu43:


chubei (approved):
Haha, a bug I didn't catch either

PR #1691, opened 2023-06-28
feature: Store app id to context file

karolisg:
This PR contains changes for cloud commands:
- During first deployment of app, `.dozer-cloud` file is created and app id is stored there
- Context app id can be changed with `dozer cloud app use {app_id}`
- All commands reads app id from context file instead of command args

### DIFF
--- a/dozer-cli/src/cli/cloud.rs
+++ b/dozer-cli/src/cli/cloud.rs
@@ -24,10 +24,10 @@ pub enum CloudCommands {
     Deploy(DeployCommandArgs),
     /// Update existing application on Dozer Cloud
     Update(UpdateCommandArgs),
-    Delete(AppCommand),
+    Delete,
     List(ListCommandArgs),
-    Status(AppCommand),
-    Monitor(AppCommand),
+    Status,
+    Monitor,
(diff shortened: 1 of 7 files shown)


dozer-cli/src/cli/cloud.rs
```diff
@@ -40,6 +40,9 @@ pub enum CloudCommands {
     /// Dozer app secrets management
     #[command(subcommand)]
     Secrets(SecretsCommand),
+    /// Dozer app secrets management
```
chubei, line 43: Wrong doc string

dozer-cli/src/lib.rs
```diff
@@ -72,6 +75,7 @@ pub fn wrapped_statement_to_pipeline(sql: &str) -> Result<QueryContext, Pipeline
     statement_to_pipeline(sql, &mut pipeline, None)
 }
 
+use crate::cli::cloud::AppCommand;
```
chubei, line 78: Need to be under cloud feature?

chubei (changes requested):

chubei (approved):

Issue #1547, opened 2023-05-09
Handle endpoint REST path collision properly


chubei:
Currently we don't do any sanity check on `endpoint.path`. If one endpoint uses a path that's a prefix of a later endpoint's path, the later one is silently not served, because path matching will always match the former endpoint first.

crajcan:
I was looking into working on this so I wrote a test to make sure we're on the same page: 

https://github.com/getdozer/dozer/pull/1693

chubei:
Hi @crajcan! Thank you for looking into this!

I checked your test and that's exactly the problem we're trying to fix. Do you mind sharing your thoughts on how you plan to tackle this?

crajcan:
> Hi @crajcan! Thank you for looking into this!
> 
> I checked your test and that's exactly the problem we're trying to fix. Do you mind sharing your thoughts on how you plan to tackle this?

Awesome!

The first thought I had was for a somewhat naive solution. The `cache_endpoints` would be reverse sorted by the length of the `path` before the respective `Scope` would be attached to the `app`. This would allow a path like `/users/posts` to be matched before `/users`.

The second implementation I thought of would be to build a tree out of the `cache_endpoint` paths. A path like `/users/posts` would become a child node of a node for `/users`. This way we could traverse the tree to create a `Scope` for the lowest nodes first, attach the default routes (`/count`, `/query`, `phase`, etc...) to that `Scope`, then attach the `Scope` to its parent `Scope` once the parent `Scope` is built. Once all of the root level `Scope`s are built, they would be added to the `app`.

The second solution might play better with future adjustments the routing logic, and if actix provides a way to generate an index of the available routes, it might provide a prettier output. But I'm not sure if it's necessary otherwise. 

Let me know if you had any other thoughts.

chubei:
Either solution sounds good to me!

PR #1693, opened 2023-06-28
fix: Handle rest path collision

crajcan:
https://github.com/getdozer/dozer/issues/1547

Reverse sorts rest endpoints before attaching to app so that overlapping routes are not incorrectly matched.

### DIFF
--- a/dozer-api/src/rest/mod.rs
+++ b/dozer-api/src/rest/mod.rs
@@ -81,7 +81,7 @@ impl ApiServer {
     fn create_app_entry(
         security: Option<ApiSecurity>,
         cors: CorsOptions,
-        cache_endpoints: Vec<Arc<CacheEndpoint>>,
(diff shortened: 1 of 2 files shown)


chubei (approved):
Hi @crajcan, thank you for your PR!

This PR looks good to me. If you can rebase and run `cargo fmt` to pass the CI, I think we're good to go.

chubei:
A unit end to end test is failing. Didn't see that coming. @crajcan do you want to take a look?

crajcan:
> A unit end to end test is failing. Didn't see that coming. @crajcan do you want to take a look?

Sure thing

chubei:
Actually it looks like a problem on `main`. My other PRs get the same error. Let me check.

chubei:
I rebased and it passed. Thanks!

PR #1697, opened 2023-06-30
fix: Remove secrets value and add secrets support in deploy command

karolisg:


dozer-cli/src/cli/cloud.rs
```diff
@@ -50,6 +53,9 @@ pub struct DeployCommandArgs {
     /// Number of replicas to serve Dozer APIs
     #[arg(short, long)]
     pub num_replicas: Option<i32>,
+
+    #[arg(short, value_parser = parse_key_val)]
+    pub secrets: Option<Vec<Secret>>,
```
chubei, line 58: `Vec` will do

dozer-cli/src/cli/cloud.rs
```diff
@@ -61,6 +67,9 @@ pub struct UpdateCommandArgs {
...
```
chubei, line 72: Same as above

chubei (approved):

PR #1712, opened 2023-07-05
fix: wildcard test case

chloeminkyung:
added sql unit test for wildcard & qualified wildcard


mediuminvader (approved):

chubei:
I can't remember what this is for. @chloeminkyung ?

PR #1718, opened 2023-07-07
feat: Support read of configs from multiple files

karolisg:
With these changes, we support reading from multiple files. By default, it reads based on two patterns: `dozer-config.*`, `queries/*.sql`

Additionally, user can provide patterns with `-c` parameter, e.g. `-c ./dozer-config.yaml -c ./dozer-config.sources.yaml -c ./queries/simple.sql`

### DIFF
--- a/dozer-cli/src/cli/helper.rs
+++ b/dozer-cli/src/cli/helper.rs
@@ -3,6 +3,7 @@ use crate::simple::SimpleOrchestrator as Dozer;
 use crate::{errors::CliError, Orchestrator};
 
 use crate::config_helper::combine_config;
+use crate::errors::CliError::ConfigurationFilePathNotProvided;
 use dozer_types::models::app_config::default_cache_max_map_size;
 use dozer_types::prettytable::{row, Table};
 use dozer_types::{models::app_config::Config, serde_yaml};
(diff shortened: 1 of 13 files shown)


dozer-cli/src/cli/helper.rs
```diff
@@ -116,9 +124,11 @@ fn parse_config(config_template: &str) -> Result<Config, CliError> {
     let config: Config = serde_yaml::from_str(&config_str)
         .map_err(|e: serde_yaml::Error| CliError::FailedToParseYaml(Box::new(e)))?;
 
-    // Create home_dir if not exists.
-    fs::create_dir_all(&config.home_dir)
-        .map_err(|e| CliError::FileSystem(config.home_dir.clone().into(), e))?;
+    if prepare_environment_fs {
```
chubei, line 127: I think this block can be removed. `dozer build` will create the directory anyway.

dozer-cli/src/config_helper.rs
```diff
@@ -0,0 +1,92 @@
...
+pub fn combine_config(config_path: &str) -> Result<String, ConfigCombineError> {
+    let mut combined_yaml = serde_yaml::Value::Mapping(Mapping::new());
+    let mut sqls = vec![];
+    let patterns: Vec<_> = config_path.split(',').collect();
+    for pattern in patterns {
+        let files_glob = glob(pattern).map_err(WrongPatternOfConfigFilesGlob)?;
+
+        for entry in files_glob {
+            let path = entry.map_err(CannotReadFile)?;
+            if let Some(name) = path.clone().to_str() {
```
chubei, line 20: We should warn about non-utf8 paths

dozer-cli/src/config_helper.rs
```diff
@@ -0,0 +1,92 @@
...
+                let content =
+                    fs::read_to_string(path.clone()).map_err(|e| CannotReadConfig(path, e))?;
+
+                if name.contains(".yml") || name.contains(".yaml") {
```
chubei, line 24: We should warn about unknown extension

chubei (approved):

chubei (approved):

PR #1722, opened 2023-07-08
feat: Serve replication log on internal grpc server

chubei:
This PR changes how app and api servers communicate. Before this PR, app writes both code (schema, generated protobuf, etc.) and data (the output log) to filesystem, and api reads them from the filesystem. Log replication is realized by api server tailing the file.

This poses a serious limitation to Dozer: app and api must share a filesystem. The only obvious solution was NFS.

In this PR, we rewrite the "internal server" that's used for app api communication. Now the internal server has two sets of APIs:

- `describe_build`: returns code that's being used by this app server
- `describe_storage` and `get_log`: returns data that's being generated by this app server.

The log replication is now realized using a "Storage" and bidirectional gRPC streaming. The mechanism is described in detail in the [Wiki](https://github.com/getdozer/dozer/wiki/Data-Stream-Replication).

The storage can be arbitrary persistent storage, NFS, FTP server, object storage service such as S3 are all possible options. In this PR we provide the `LocalStorage` implementation so Dozer can still run locally or over NFS.

Other changes include:

- Rename some "migration" to "build"
- Remove a `unwrap` in `dozer-api/src/grpc/client_server.rs`
- Add a config section `app` which includes some old top level options and new options related to new log reader


snork-alt:
@chubei did you check out OpenDAL (https://github.com/apache/incubator-opendal) for storage access? Might be useful

chubei:
> @chubei did you check out OpenDAL (https://github.com/apache/incubator-opendal) for storage access? Might be useful

Looks like a decent library. We can switch to that in the future to support more storage out of the box.

mediuminvader (approved):

v3g42 (approved):

PR #1723, opened 2023-07-10
chore: set bucket for histogram

duonganhthu43:


dozer-tracing/src/telemetry.rs
```diff
@@ -29,7 +29,16 @@ pub fn init_telemetry(app_name: Option<&str>, telemetry_config: Option<Telemetry
 
     if let Some(telemetry_config) = telemetry_config {
         if telemetry_config.metrics.is_some() {
-            PrometheusBuilder::new()
+            let mut prom_builder = PrometheusBuilder::new();
+            prom_builder = prom_builder
```
chubei, line 33: The bucket value range is scary. Let's use https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.set_buckets_for_metric

chubei (approved):

PR #1728, opened 2023-07-10
fix: Use full record as a join hashmap key if there is no pk

mediuminvader:


chubei (approved):

chubei:
We need a test that prevents this bug from happening again.

chubei:
@mediuminvader let's wrap this up?

chubei:
Turns out if there's no PK all fields are used.

PR #1736, opened 2023-07-11
feat: Implement config override

chubei:
We introduce a new repeated command line parameter: `--config-overrides`.

Example:

```bash
dozer --config-overrides /app.commit_size=100
```

The part before `=` is a JSON pointer, and the part after `=` should be a valid JSON string. If the JSON pointer points to an existing config value, the JSON value will be replaced.

# Breaking change

This feature requires us to remove `!Ref` in `source.connection`, because it's implemented using custom deserialization, and doesn't work with JSON. All configuration files should be updated accordingly.

We don't allow using null as empty sequence anymore. For example:

```yaml
columns:
```

and

```yaml
index:
  primary_key:
```

`app_name` must be present.


karolisg (approved):
@mediuminvader We need to check all examples and samples, because of that change in config

PR #1742, opened 2023-07-12
fix: Use different slot name everytime when dozer connects to postgres

karolisg:


v3g42:
@karolisg i would also give an option to definite a slot name instead. So it is UpTo the user.

v3g42:
I m talking about reusing a slot that is already created. Let's say if a replication slot is already present. 

But that could also be handled in a different PR.

PR #1744, opened 2023-07-13
fix: synchronize schema change

duonganhthu43:


dozer-types/protos/cloud_types.proto
```diff
@@ -159,8 +154,9 @@ message Source {
   string name = 1;
   string table_name = 2;
   repeated string columns = 3;
-  Connection connection = 4;
-  RefreshConfig refresh_config = 5;
+  string connection = 4;
+  string schema = 5;
```
chubei, line 158: This is optional

chubei (approved):
Tag 12 (field app) is missing. I guess it doesn't matter?

PR #1746, opened 2023-07-13
fix: Dont serialize app name if it is missing

karolisg:


duonganhthu43 (approved):

dozer-sql/src/pipeline/projection/factory.rs
```diff
@@ -73,16 +73,14 @@ impl ProcessorFactory<SchemaSQLContext> for ProjectionProcessorFactory {
                         })
                         .collect();
                     for f in fields {
-                        let res = parse_sql_select_item(&f, input_schema);
-                        if let Ok(..) = res {
-                            select_expr.push(res.unwrap())
+                        if let Ok(res) = parse_sql_select_item(&f, input_schema) {
```
chubei, line 76: @chloeminkyung we're just ignoring any select that fails to parse? This doesn't sound right.

PR #1747, opened 2023-07-14
Pin rust version

chloeminkyung:
pin Rustup toolchain version to `RUSTUP_TOOLCHAIN_VERSION: 1.25.2` instead of `stable`


chubei:
Just push to `release-dev` to test

PR #1756, opened 2023-07-18
feat: Show correct error when organisation doesnt exist

karolisg:


dozer-cli/src/main.rs
```diff
@@ -187,25 +188,44 @@ fn run() -> Result<(), OrchestrationError> {
...
-                    info!("Using \"{app_id}\" app");
-                    Ok(())
+            Commands::Cloud(cloud) => {
+                render_logo();
+
+                match cloud.command.clone() {
+                    CloudCommands::Deploy(deploy) => dozer.deploy(cloud, deploy),
+                    CloudCommands::Api(api) => dozer.api(cloud, api),
+                    CloudCommands::Login { organisation_name } => {
+                        info!("Organisation and client details can be created in https://dashboard.dev.getdozer.io/login \n");
```
chubei, line 198: Should we move all these into `login`?

chubei (approved):

PR #1758, opened 2023-07-18
feat: Use config argument in deploy command

karolisg:


chubei (approved):

chubei:
I've always wanted this. Cool!

PR #1773, opened 2023-07-20
fix: Fix config combination with separated sql scripts

karolisg:


dozer-cli/src/config_helper.rs
```diff
@@ -67,16 +46,39 @@ pub fn combine_config(config_paths: Vec<String>) -> Result<Option<String>, Confi
...
+        warn!("Config file \"{name}\" extension not supported");
+    }
+
+    Ok(())
+}
+
+pub fn merge_yaml(
     from: serde_yaml::Value,
     to: &mut serde_yaml::Value,
+    join_strings: bool,
```
chubei, line 74: This parameter is only for the branch `name.contains(".sql")`. We don't really need to call `merge_yaml` there. Just insert or append to the `sql` string.

karolisg: Sure. Refactored that part.

dozer-cli/src/config_helper.rs
```diff
@@ -67,7 +46,38 @@ pub fn combine_config(config_paths: Vec<String>) -> Result<Option<String>, Confi
...
+            .map_err(|e| ConfigCombineError::ParseYaml(name.to_string(), e))?;
+        merge_yaml(yaml, combined_yaml)?;
+    } else if name.contains(".sql") {
+        combined_yaml
+            .as_mapping_mut()
+            .expect("Should be mapping")
+            .entry(serde_yaml::Value::String("sql".into()))
+            .and_modify(|s| {
+                let query = s.as_str();
+                *s = match query {
```
chubei, line 65: We should return an error here? The value of `sql` should always be string. If user writes something else, we can let them know by returning error instead of silently override?

karolisg: Right. Changed to error result.

chubei (approved):

PR #1800, opened 2023-07-27
feat: implement nested connector

v3g42:
Dozer apps can be now stacked like lego blocks.

https://github.com/getdozer/dozer-samples/pull/69
```yaml
connections:
  - config: !Dozer
      grpc:
        host: 0.0.0.0
        ## Default port of upstream dozer internal grpc server
        port: 50053
    name: nested_dozer

```


chubei:
Replaced by #1883

PR #1802, opened 2023-07-28
fix: custom enum row detection, support different type Coalesce

chloeminkyung:
fix: custom enum row detection, support different type Coalesce

### DIFF
--- a/dozer-ingestion/src/connectors/postgres/schema/helper.rs
+++ b/dozer-ingestion/src/connectors/postgres/schema/helper.rs
@@ -321,15 +321,15 @@ impl SchemaHelper {
         let type_oid: u32 = row.get(6);
 
         // TODO: workaround - in case of custom enum
-        let typ;     if type_oid == 28862 {
-            typ = FieldType::String;
+        let typ = if type_oid == 28862 {
+            FieldType::String
         } else {
             let oid_typ = Type::from_oid(type_oid);
-            typ = oid_typ.map_or_else(
+            oid_typ.map_or_else(
                 || Err(InvalidColumnType(column_name.clone())),
                 postgres_type_to_dozer_type,
-            )?;
-        }
+            )?
+        };
 
         let replication_type = String::from_utf8(vec![replication_type_int as u8])
             .map_err(|_e| ValueConversionError("Replication type".to_string()))?;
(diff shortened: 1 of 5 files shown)


dozer-ingestion/src/connectors/postgres/schema/helper.rs
```diff
@@ -319,12 +319,17 @@ impl SchemaHelper {
...
         let replication_type_int: i8 = row.get(5);
         let type_oid: u32 = row.get(6);
-        let typ = Type::from_oid(type_oid);
 
-        let typ = typ.map_or_else(
-            || Err(InvalidColumnType(column_name.clone())),
-            postgres_type_to_dozer_type,
-        )?;
+        // TODO: workaround - in case of custom enum
+        let typ = if type_oid == 28862 {
```
chubei, line 324: I think the same workaround is needed in `xlog_mapper` otherwise streaming won't work

chubei (commented):
1. Add the newly supported type to the postgres connector test?

dozer-ingestion/src/connectors/postgres/schema/helper.rs
chloeminkyung: gotcha

chloeminkyung:
```rust
            // TODO: workaround - in case of custom enum
            let type_oid = column.type_id() as u32;
            let typ = if type_oid == 28862 {
                Type::VARCHAR
            } else {
                Type::from_oid(type_oid).ok_or_else(|| {
                    PostgresSchemaError::InvalidColumnType(column_name.to_string())
                })?
            };

            columns.push(TableColumn {
                name: column_name.to_string(),
                flags: column.flags(),
                r#type: typ,
                column_index,
            })
```
this would be the corresponding changes in the `xlog_mapper`, let me know if this looks good

chubei: Write test and find out

chloeminkyung: Just confirming before we move on to the testing

chubei: Looks good to me, but I'm not sure without testing

mediuminvader (approved):

Issue #1666, opened 2023-06-22
MongoDB connector


snork-alt:
## Description
Implement MongoDB connector for Dozer

## Details
The connector should use MongoDB CDC similarly to the behavior of the PostgreSQL connector. All data changes from MongoDB should be translated into Dozer Inserts, Deletes, or Updates. As MongoDB is a schemaless database the ingested record should contain a single column of JSON type.

## Deliverables
The implementation should come with comprehensive tests and a sample to be pushed to `dozer-samples` repo

Jesse-Bakker:
@snork-alt While reading the postgres iterator/snapshotter implementation, I noticed that snapshotting happens in a different connection than the creation of the replication slot. I this means that, if a write transaction commits between starting the transaction that creates the replication slot and the start of a snapshotting transaction, a row might be, for example, inserted twice or deleted twice.

Am I understanding replication slots and transaction isolation incorrectly or is this a bug?

snork-alt:
@karolisg and @chubei can provide more details

Jesse-Bakker:
In mongo, there are the concepts of "databases" and "collections". Collections map cleanly to tables, but databases can map to either schemas or separate connections. 
What would you say is the preferred way to go here?

snork-alt:
@Jesse-Bakker I'd say schemas as a connection should be mapped to different physical db connection. @karolisg @chubei what's your opinion?

chubei:
I'll prefer mapping mongodb databases to connections, similar to how our postgres connector always connects to one database.

snork-alt:
Let's follow @chubei suggestion. He is the boss 😁

universalmind303:
> As MongoDB is a schemaless database the ingested record should contain a single column of JSON type.

So I'm wondering if we should infer a schema instead. Using JSON will likely result in data loss as BSON doesn't directly translate to JSON.

snork-alt:
That's a good point @universalmind303. I think we should give an option. However, there is no guarantee that schema will be the same for all objects in the same collection. How shall we handle this ?

Jesse-Bakker:
I think imposing a schema on a schemaless database will add complexity and add failure modes and will only be beneficial for the small number of cases where using mongodb wasn't the right choice in the first place. 

I think if the goal is to maintain fidelity of the data, the two main options would be to either

 - Extend the JSON data type in dozer to be more like BSON
Or
[...]

PR #1810, opened 2023-07-31
Add Mongodb connector MVP

Jesse-Bakker:
Initial implementation of a mongodb connector.

This implementation uses the `changeStream` API to replicate changes to mongodb collections. Consistency is achieved by snapshotting in a session with `snapshot` read concern, storing the operation time of the snapshot read and starting the changeStream from that timestamp.

There are some restrictions on the mongo side:

 1. The collection should have pre- and post-images enabled (minimum mongodb version 6.0)
 2. The mongodb server cannot be sharded

Reasoning:
1: By default, mongo change stream events for updates only contain a delta between the previous document and the updated document. The dozer pipeline does not support delta's for json columns, meaning the entire updated document should be passed into the pipeline. If the full document is retrieved using the `"fullDocument": "lookup"` option, it is queried by the server after the change event is emitted, allowing transactional consistency to be broken.
2.: Non-sharded collections always have a unique index on the `_id` field. In sharded collections, there might be a unique index on the sharding key, but this is not the default. Checking this would add complexity, so i left it out for the initial implementation, disallowing sharded servers entirely.

Furthermore, the current test harness is not really suitable for key-value sources (witnessed by the fact that there are no tests for the kafka and ethereum connectors. I feel setting up a new test harness is not really in-scope for this feature, but when it is in place, it should be fairly easy to add more tests.

To do:
wire up the connector configuration 

/claim #1666


chubei:
Hi @Jesse-Bakker, thank you for your work! From what I can see, this is generally in good shape. Regarding the two problems:

1. Dozer actually keeps track of inserted rows, in case the data source doesn't give full row data on deletes and updates. Do you think it's possible for the mongodb connector to reconstruct the deleted and updated document if it has access to original data?
2. I don't fully get the complexity related to sharding. If it's just one `if` away, maybe we can include it in this PR.

Jesse-Bakker:
@chubei

1. Yes, if the connector has access to the original data, it can reconstruct the updated document. Deletes are no problem, as they would just need the primary key, which is always present in change events
2. Supporting sharded servers is not too hard, but if a collection is sharded, the following needs to happen:
 - Find whether the sharded collection is unique on the shard key, and what the shard key is (requires reading from the config database, which does not have a stable format)
 - If the shard key is not unique, error out (no available unique primary key), or else
 - Construct the shard key manually for the existing documents during snapshotting (change events contain it in the `document_key` field)

chubei:
Sounds like supporting sharded collections is too much work.

snork-alt:
Hey @Jesse-Bakker! Thanks for your contribution! Are you up for a chat? Please ping me at matteo@getdozer.io

Jesse-Bakker:
For some reason, github actions refuses to re-parse the workflow file

chubei:
> For some reason, github actions refuses to re-parse the workflow file

I think that's because we're using the pull request target trigger.

chubei (approved):

PR #1829, opened 2023-08-07
feat: Introduce MySQL Connector

### LINKED ISSUE #1670: MySQL Connector
## Description
Implement MySQL connector for Dozer

## Details
The connector should use MySQL CDC similarly to the behavior of the PostgreSQL connector. All data changes from MySQL should be translated into Dozer Inserts, Deletes, or Updates. Connector should support resuming. 

## Deliverables
The implementation should come with comprehensive tests and a sample to be pushed to `dozer-samples` repo

abcpro1:
Define and implement a MySQL connector for Dozer. 

It operates similarly to the Postgres connector; especially when it comes to listing tables, columns, and schemas.

For replication, MySQL CDC is used to replay database changes in Dozer, after initially copying the table contents with a simple SELECT query.

The connector implementation is pretty much complete with the only exception of partial updates to JSON fields, which is a feature in MySQL that does not have pre-existing support in Dozer as far as I could tell.

closes #1670
/claim #1670

### DIFF
--- a/dozer-ingestion/Cargo.toml
+++ b/dozer-ingestion/Cargo.toml
@@ -45,6 +45,10 @@ tokio-postgres-rustls = "0.10.0"
 rustls-native-certs = "0.6.2"
 rand = "0.8.5"
 url = "2.4.0"
+mysql_async = { version = "0.32.2", default-features = false, features = ["default-rustls"] }
+mysql_common = { version = "0.30", default-features = false, features = ["chrono", "rust_decimal"] }
+chrono = "0.4.26"
+geozero = { version = "0.10.0", default-features = false, features = ["with-wkb"] }
 
 [dev-dependencies]
 criterion = { version = "0.4.0", features = ["html_reports"] }
--- a/dozer-ingestion/src/connectors/mod.rs
+++ b/dozer-ingestion/src/connectors/mod.rs
@@ -3,6 +3,7 @@ pub mod ethereum;
 pub mod grpc;
 #[cfg(feature = "kafka")]
 pub mod kafka;
+pub mod mysql;
 pub mod object_store;
 pub mod postgres;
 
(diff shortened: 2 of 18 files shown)


snork-alt:
Thanks for your contribution @abcpro1. Waiting for @chubei to review so we can merge and payout the bounty. @abcpro1 would u be up for a call ? Can drop me a note at matteo@getdozer.io ?

chubei:
Hi @abcpro1 thank you for your work! The implementation looks amazing and I can't wait to merge it.

I hope you can add several more things to this PR. Please let me know what you think.

1. Implement `MySqlConnector::types_mapping`.

This method is mostly for documentation purpose, so we can quickly see which types are supported instead of digging into the code.

2. Add an integration test in `dozer-ingestion/tests`.

I see that you've added an end to end test in `dozer-tests` and really appreciate that. It's not documented by we also have an integration test suite in `dozer-ingestion/tests` specifically designed for testing the connectors (while `dozer-tests` tests the whole dozer in an end to end fashion).

To add a new connector to the integration test suite is very easy. You implement [several traits](https://github.com/getdozer/dozer/blob/main/dozer-ingestion/tests/test_suite/mod.rs) which does the database setup, and [add a test](https://github.com/getdozer/dozer/blob/main/dozer-ingestion/tests/test_connectors.rs#L14) that calls the test suite on your connector.

You can take the `PostgresConnectorTest` as a reference.

abcpro1:
@chubei Thank you for reviewing this pull request. I'm glad you like it. I will add comprehensive tests as you suggested. As for `MySqlConnector::types_mapping`, I have a question: does "external type" here mean MySQL type?

chubei:
> does "external type" here mean MySQL type?

Yes.

snork-alt:
@abcpro1 it would be great if you can also document in the PR all the config parameters of the connector and maybe add a small sample using MySQL in `dozer-samples`. Thanks.

abcpro1:
@snork-alt Sure.

chubei (approved):

PR #1850, opened 2023-08-14
fix: Mark log creation to avoid writing to old log version

chubei:
Bug happens in following senario:

1. App and api both runs and one operation is ingested. At this time, the operation is not persisted to log, because there's only one operation and it doesn't reach the persisting threshold. The operation is persisted to cache because api gets the operation from app gRPC stream.
2. App and api restarts and another operation is ingested. This operation will be seen by the app, but api will ignore this operation, because app believes this is the first operation, while api is only interested in the second operation and so on.

We fix this by including a "marker" file in the log directory, so app knows that the log may contain unpersisted operations and create a new log.

This is a temporary workaround. We need to solve app/api persistence mismatch properly during pipeline checkpointing.

### DIFF
--- a/dozer-log/src/replication/mod.rs
+++ b/dozer-log/src/replication/mod.rs
@@ -3,6 +3,7 @@ use std::ops::{DerefMut, Range};
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
+use camino::Utf8Path;
 use dozer_types::grpc_types::internal::storage_response;
 use dozer_types::log::{debug, error};
 use dozer_types::models::app_config::LogStorage;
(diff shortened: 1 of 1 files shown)


dozer-log/src/replication/mod.rs
```diff
@@ -93,12 +94,23 @@ impl Log {
     }
 
     pub async fn new(options: LogOptions, log_dir: String, readonly: bool) -> Result<Self, Error> {
-        let (storage, prefix) = create_log_storage(options.storage_config, log_dir).await?;
+        let (storage, mut prefix) = create_log_storage(options.storage_config, log_dir).await?;
+        if !readonly {
```
Jesse-Bakker, line 98: Can you add a comment here explaining what the marker file does, with a `TODO:` or `XXX:` tag or something to revisit this in the future?

Jesse-Bakker (approved):
LGTM. Left a small suggestion, but feel free to merge

dozer-log/src/replication/mod.rs
chubei: Added. I'll resolve and merge. Thanks!

PR #1851, opened 2023-08-14
error for SELECT statements with no INTO clause

### LINKED ISSUE #1832: `dozer-sql` allows `SELECT` statement without `INTO` clause
We only checked [if there's at least one `INTO`](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/builder.rs#L106), but with multiple statements, if one has `INTO` and others don't, we don't report error.

aaryaattrey:
Closes #1832


dozer-sql/src/pipeline/builder.rs
```diff
@@ -117,6 +118,7 @@ fn query_to_pipeline(
     query_ctx: &mut QueryContext,
     stateful: bool,
     pipeline_idx: usize,
+    is_top_select: &mut bool,
```
chubei, line 121: `is_top_select: bool` should be enough? This doesn't look like a input output parameter.

dozer-sql/src/pipeline/builder.rs
chubei, line 1: We should be able to remove the check and error variant in https://github.com/getdozer/dozer/pull/1851/files#diff-48e511c9d5fa40c682baa4c550d5f368ee49889bff0f70e3f352fb5d29a39234R107 ?

chubei (commented):
Looks great. Two style related comments.

dozer-sql/src/pipeline/builder.rs
Jesse-Bakker: Maybe change it to an assert to sanity check?

aaryaattrey: Yes, we can remove it.

> We should be able to remove the check and error variant in https://github.com/getdozer/dozer/pull/1851/files#diff-48e511c9d5fa40c682baa4c550d5f368ee49889bff0f70e3f352fb5d29a39234R107 ?

aaryaattrey: > Maybe change it to an assert to sanity check?

I think returning an error is more helpful to the user than an assertion.

chubei: > Maybe change it to an assert to sanity check?
It's OK to remove. We added checks at earlier place.

dozer-sql/src/pipeline/builder.rs
```diff
@@ -690,3 +705,44 @@ mod tests {
...
+
+#[test]
+fn test_missing_into_clause() {
+    let sql = r#"
+    SELECT
+        a.name as "Genre",
+        SUM(amount) as "Gross Revenue(in $)"
+    INTO gross_revenue_stats
+    FROM
+    (
```
Jesse-Bakker, line 717: The check might be triggering on this inner query instead of the outer one. I would suggest splitting this test into multiple simpler ones which each test one edge case. For example:

```sql
SELECT a INTO c FROM b
```
**should not** trigger an error

```sql
SELECT a FROM b
```

**should** trigger an error

```sql
SELECT a INTO c FROM (SELECT a from b)
```

**should not** trigger an error

```sql
SELECT a FROM (SELECT a FROM b)
```

**should** trigger an error

dozer-sql/src/pipeline/errors.rs
```diff
@@ -84,7 +82,8 @@ pub enum PipelineError {
     IllegalFieldIdentifier(String),
     #[error("Unable to cast {0} to {1}")]
     UnableToCast(String, String),
-
+    #[error("Each SQL Select Statement must have a corresponding INTO clause")]
```
Jesse-Bakker, line 85: Let's make the error message a bit more in the style of the others. So, instead of stating the desired outcome, state what went wrong, e.g. `missing INTO clause for top-level SELECT statement`

dozer-sql/src/pipeline/builder.rs
```diff
@@ -690,3 +705,65 @@ mod tests {
...
+fn test_missing_into_in_with_clause() {
+    let sql = r#"WITH tbl as (select a from B)
+    select B
+    from tbl;"#;
+    let result = statement_to_pipeline(sql, &mut AppPipeline::new(), None);
+    //check if the result is an error
+    assert!(result.is_err());
+    //check if the error is of type MissingIntoClause
+    let error = result.unwrap_err();
+    assert!(matches!(error, PipelineError::MissingIntoClause));
```
Jesse-Bakker, line 757: You can use [pattern matching](https://doc.rust-lang.org/book/ch18-00-patterns.html) here to simplify this.

```rust
assert!(matches!(result, Err(PipelineError::MissingIntoClause)))
```

The `matches!(result, Err(PipelineError::MissingIntoClause))` macro is equivalent to something like
```rust
match result {
    Err(PipelineError::MissingIntoClause) => true,
    _ => false,
}
```

Jesse-Bakker (commented):
Looks very good! 
I left a comment about the error message, which is mostly about consistency and user experience. 
The other comment is just a nice learning opportunity/stylistic comment.

dozer-sql/src/pipeline/builder.rs
chubei: I mean we can write `assert!(matches!(result, Err(PipelineError::MissingIntoClause))`;

Jesse-Bakker: Yes, I meant to say to use the macro. Just wanted to explain what the macro meant

chubei: Ah sorry, I misread your comment

Jesse-Bakker (approved):

PR #1855, opened 2023-08-15
chore: Run CI on default runner

mediuminvader:


.github/workflows/coverage.yaml
chubei, line 1: Runner is not changed?

.github/workflows/e2e.yaml
```diff
@@ -1,9 +1,8 @@
 name: Dozer E2E Test
 
 on:
-  pull_request_target:
-    branches: [main]
-  workflow_dispatch:
+  schedule:
```
chubei, line 4: Don't need to run on `merge_group` if we're running nightly? And `workflow_dispatch` can be kept so it's easier to test

chubei:
`ci.yaml` runner is not changed?

.github/workflows/coverage.yaml
chubei, line 1: Does `group: Org Larger Runners` work?

.github/workflows/dozer.yaml
chubei, line 1: Does `group: Org Larger Runners` work?

chubei (approved):
I don't know if group affects anything. How did you test?

.github/workflows/dozer.yaml
mediuminvader: should just remove, right?

chubei: Probably. Have to test

chubei:
@mediuminvader are you still working on it? `coverage.yaml` and the workflows in `dozer-cloud` are still using larger runners.

chubei:
Replaced by #1910

PR #1862, opened 2023-08-16
feat: replace jaeger with xray by using ADOT

duonganhthu43:
https://aws-otel.github.io/docs/introduction

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1143,9 +1143,9 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.6.7"
+version = "0.6.11"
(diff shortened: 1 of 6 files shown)


chubei (approved):
Does all clusters talk to the same ADOT or we have one ADOT per region?

duonganhthu43:
> Does all clusters talk to the same ADOT or we have one ADOT per region?

Right now for testing we just use same ADOT and differentiate by `service.name` aka namespace => All trace and error will be in same region 
Depend on how we want to distribute the ERRORs based on region or unify everything in same place we could adjust the structure

PR #1867, opened 2023-08-19
test: unittests for the MySQL connector

abcpro1:
Increase the test coverage of the MySQL connector with a combination of unittests and integration tests.

### DIFF
--- a/dozer-ingestion/src/connectors/mysql/binlog.rs
+++ b/dozer-ingestion/src/connectors/mysql/binlog.rs
@@ -184,6 +184,7 @@ impl BinlogIngestor<'_, '_, '_, '_, '_> {
                                 *self.txn, seq_no,
                             ))
                             .map_err(ConnectorError::IngestorError)?;
+                        seq_no += 1;
                     }
                 }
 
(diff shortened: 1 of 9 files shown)


dozer-ingestion/src/connectors/mysql/conversion.rs
```diff
@@ -190,3 +187,202 @@ impl GeomProcessor for PointProcessor {
...
+    use chrono::{DateTime, NaiveDate, NaiveDateTime, NaiveTime, Offset, Utc};
+    use dozer_types::{
+        json_types::JsonValue,
+        types::{DozerDuration, Field, FieldType, TimeUnit},
+    };
+    use mysql_async::Column;
+    use mysql_common::{constants::ColumnType, row::new_row, Value};
+
+    #[test]
+    fn test_field_types() {
```
Jesse-Bakker, line 207: I think it might be more useful to extract the parsing logic for data types into a separate function and test that (so, test that "decimal(5, 2)" parses to "decimal"), instead of basically replicating the match statement in that function in the test.

A metric I like to use for deciding if something is worth testing, is if the test would be of a lower order of complexity than the thing that's being tested. In this case, most of the test tests a simple match statement and therefore you are just as likely to make a mistake writing the test as you are writing the implementation.

dozer-ingestion/src/connectors/mysql/binlog.rs
```diff
@@ -655,3 +656,174 @@ impl<'a> BinlogRowsEvent<'a> {
...
+        constants::ColumnType,
+        io::ParseBuf,
+        Value,
+    };
+
+    use crate::connectors::mysql::conversion::{IntoField, IntoFields};
+
+    #[test]
+    fn test_row_conversion() {
+        use ColumnType::*;
```
Jesse-Bakker, line 684: For this test, it's very hard to see what is actually compared to what, and therefore, I can't see what is being tested. Ideally, you'd have a function
```rust
fn check(binlog_value: BinlogValue, column: Colum, expected: Field) {
// Some asserts
}
```
and the test just looks like
```rust
check(Jsonb(Null), MYSQL_TYPE_JSON, JsonValue::Null);
check(Jsonb(I16(2)), MYSQL_TYPE_JSON, JsonValue::Number(2.0.into()))
```
There, you can directly see that the test is correct.

The same goes for some of the other tests

abcpro1: Agreed. Let me change it.

dozer-ingestion/src/connectors/mysql/conversion.rs
Jesse-Bakker: I don't think changing how it works would make it harder to understand (the structure of the test would be the same, just with slightly different data).

I agree it isn't worth changing it now, as you have already done the work, the values won't change, and it's definitely not a bad thing to be thorough!

Jesse-Bakker (approved):
Great work!

dozer-ingestion/src/connectors/mysql/binlog.rs
```diff
@@ -141,7 +141,14 @@ impl BinlogIngestor<'_, '_, '_, '_, '_> {
         'binlog_read: while let Some(event) = self.binlog_stream.as_mut().unwrap().next().await {
             let binlog_event = event.map_err(MySQLConnectorError::BinlogReadError)?;
 
-            self.next_position.position = binlog_event.header().log_pos().into();
+            let is_artificial = binlog_event
```
Jesse-Bakker, line 144: Should the binlog still be rotated on line 180 if the rotate event is artificial? If not, just doing `if is_artificial { continue; }` might more clearly signal the intent that this event is just skipped.

dozer-ingestion/src/connectors/mysql/binlog.rs
```diff
@@ -139,8 +139,24 @@ impl BinlogIngestor<'_, '_, '_, '_, '_> {
         let mut table_cache = BinlogTableCache::new(self.tables);
 
         'binlog_read: while let Some(event) = self.binlog_stream.as_mut().unwrap().next().await {
+            match self.local_stop_position {
```
Jesse-Bakker, line 142: Moving this block to above `self.next_position.position = binlog_event.header().log_pos().into();` changes the behavior. Is that intended?

abcpro1: Some unrelated strange test failures were happening when I added `if is_artificial { continue; }`, and they when away when I moved this block from the bottom to the top. It's a minor change and it's still correct. I am not going to try to figure out what's going on because that's valuable time that I should spend elsewhere.

Jesse-Bakker: :+1: , just wanted to make sure it was intentional

Jesse-Bakker (approved):

PR #1874, opened 2023-08-21
fix: start dozer live ui server

duonganhthu43:


chubei (approved):
Oh I didn't know we're using path to route in the SPA. I thought people still use fragments.

PR #1883, opened 2023-08-21
feat: Dozer connector

Jesse-Bakker:
Supersedes #1800


dozer-tests/src/e2e_tests/runner/running_env.rs
```diff
@@ -367,6 +367,7 @@ fn write_dozer_config_for_running_in_docker_compose(
                     .unwrap();
                 mysql.url = url.into();
             }
+            ConnectionConfig::Dozer(_) => {}
```
chubei, line 370: Let's leave a `todo!` here instead of ignoring the connector.

dozer-ingestion/src/connectors/dozer/connector.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    async fn start(
+        &self,
+        ingestor: &Ingestor,
+        tables: Vec<TableInfo>,
+    ) -> Result<(), ConnectorError> {
+        let mut joinset = JoinSet::new();
+        for (table_index, table) in tables.into_iter().enumerate() {
+            let builder = self.get_reader_builder(table.name.clone()).await?;
+            joinset.spawn(read_table(table_index, builder, ingestor.clone()));
```
chubei, line 130: We'll need to make the ingestion order deterministic because we want dozer connector to support restart. Leaving a comment here to remind myself.

dozer-log/src/errors.rs
```diff
@@ -23,4 +23,6 @@ pub enum ReaderError {
     Storage(#[from] crate::storage::Error),
     #[error("Reader thread has quit: {0:?}")]
     ReaderThreadQuit(#[source] Option<tokio::task::JoinError>),
+    #[error(transparent)]
+    ReaderBuilderError(#[from] ReaderBuilderError),
```
chubei, line 27: Remove this. I intentionally differentiate between the building phase error and running phase error.

dozer-ingestion/src/connectors/dozer/connector.rs
```diff
@@ -0,0 +1,235 @@
...
+    reader_builder: LogReaderBuilder,
+    ingestor: Ingestor,
+) -> Result<(), ConnectorError> {
+    let mut reader = reader_builder.build(0, None);
+    loop {
+        let (op, seq_no) = reader.next_op().await.map_err(|e| {
+            ConnectorError::NestedDozerConnectorError(NestedDozerConnectorError::ReaderError(e))
+        })?;
+        let msg = match op {
+            LogOperation::Op { op } => IngestionMessage::new_op(0, seq_no, table_idx, op),
```
chubei, line 222: `seq_no` has to be unique across the whole connection, as stated in https://github.com/getdozer/dozer/blob/main/dozer-ingestion/SPEC.md#ingestionmessage.

dozer-ingestion/src/connectors/dozer/connector.rs
```diff
@@ -0,0 +1,235 @@
...
+            LogOperation::Commit { .. } | LogOperation::SnapshottingDone { .. } => continue,
```
chubei, line 223: We need to think about what "snapshotting" means for dozer connector. Not sending `SnapshottingDone` is not an option because downstream is relying on this message to enable certain optimizations.

The solution in my mind is like following:

`LogOperation::SnapshottingDone` should not include `connection_name`, and is only emitted once all connection affecting this log is done snapshotting.

We can leave it for another PR.

dozer-ingestion/tests/test_suite/connectors/dozer.rs
```diff
@@ -0,0 +1,346 @@
...
+};
+
+pub struct DozerConnectorTest {
+    _tmpdir: TempDir,
+    shutdown: Option<(JoinHandle<()>, ShutdownSender)>,
+    ingest_state: Arc<Mutex<(u32, IngestServiceClient<Channel>)>>,
+    schema: String,
+}
+
+fn field_to_grpc(field: Field) -> Option<value::Value> {
```
chubei, line 42: I think we have this function somewhere. May just have to make that function public.

chubei (changes requested):

dozer-ingestion/src/connectors/dozer/connector.rs
Jesse-Bakker: I assumed here that the seq_no would be guaranteed to be unique by the log we're reading from, but that could of course come from multiple connectors, making it not unique. Thanks for spotting this!

dozer-api/src/grpc/internal/internal_pipeline_server.rs
```diff
@@ -109,6 +119,14 @@ impl InternalPipelineService for InternalPipelineServer {
     }
 }
 
+async fn get_build_response(endpoint: &LogEndpoint) -> BuildResponse {
```
chubei, line 122: We can remove the `async`?

chubei (approved):
Wonderful!

Just one `async` that can be removed. Feel free to resolve and merge.

PR #1885, opened 2023-08-21
handle-invalid-request

### LINKED ISSUE #1852: REST `query` API ignores invalid JSON
If user includes malformed JSON in the request body, it's just ignored and default query is executed. This causes confusion to users.

We need to answer several questions and define a new behaviour.

1. Do we execute query requests with no `content-type` header?
2. Do we execute query requests with `content-type` header that's not `application/json`?
3. If we do in 1 and 2, do we report malformed JSON body?
4. Does empty body count as malformed JSON?

My preference:

1. We execute query requests with no `content-type` header.
2. We don't execute query requests with `content-type` header that's not `application/json`.
3. We report malformed JSON body.
4. We don't treat empty body as malformed JSON and just execute the default query.

@getdozer/dozer-dev I'd like to hear your opinions.

aaryaattrey:
Closes   #1852


chubei (commented):
Is it possible to reuse `Json` extractor? This implementation doesn't have size limit.

chubei:
Besides we should write unit tests for the `FromRequest` implementation.

dozer-api/src/rest/tests/routes.rs
```diff
@@ -257,3 +259,112 @@ async fn path_collision_test() {
     //assert the route matched something
     assert_ne!(res.status(), 404);
 }
+
+#[actix_web::test]
```
chubei, line 263: Is it possible to test the `FromRequest` implementation without setting up the service?

dozer-api/src/rest/tests/routes.rs
```diff
@@ -257,3 +259,112 @@ async fn path_collision_test() {
...
+    let cache_manager = test_utils::initialize_cache(&endpoint.name, None);
+    let api_server = ApiServer::create_app_entry(
+        None,
+        CorsOptions::Permissive,
+        vec![Arc::new(
+            CacheEndpoint::open(&*cache_manager, Default::default(), endpoint.clone()).unwrap(),
+        )],
+        Default::default(),
+    );
+    let app = actix_web::test::init_service(api_server).await;
```
chubei, line 275: These lines are copied everywhere. Should be able to put into a function?

dozer-api/src/rest/tests/routes.rs
```diff
@@ -257,3 +259,112 @@ async fn path_collision_test() {
...
+
+    let (count, records) = count_and_query(&endpoint.path, &app, None).await;
```
chubei, line 300: Don't have to test again as it's already tested in other places?

dozer-api/src/rest/tests/routes.rs
```diff
@@ -257,3 +259,112 @@ async fn path_collision_test() {
...
+        .uri(&format!("{path}/count", path = endpoint.path))
+        .insert_header(ContentType::json())
+        .to_request();
+    let res_count = actix_web::test::call_service(&app, req_count).await;
+
+    // Assert success
+    assert!(res_count.status().is_success());
+    let body: Value = actix_web::test::read_body_json(res_count).await;
+    let count = body.as_u64().unwrap();
+    assert_eq!(count, 52);
```
chubei, line 341: We can add a `content_type` parameter to `count_and_query` to avoid code duplication?

dozer-api/src/rest/mod.rs
```diff
@@ -98,9 +99,10 @@ impl ApiServer {
             .iter()
             .map(|cache_endpoint| cache_endpoint.endpoint.path.clone())
             .collect();
-
+        let cfg = PayloadConfig::default().limit(4 * 1024);
```
chubei, line 102: Let's stick with default payload limit for now

chubei (changes requested):
Nicely done! Just some style and test related comments. Let me know what you think.

dozer-api/src/rest/tests/routes.rs
chubei: I'm just wondering if it's possible. If not testing with service is fine.

dozer-api/src/rest/tests/routes.rs
```diff
@@ -99,10 +115,10 @@ async fn count_and_query_route() {
     let app = actix_web::test::init_service(api_server).await;
```
chubei, line 115: Also call `setup_service`

dozer-api/src/rest/tests/routes.rs
```diff
@@ -63,24 +66,37 @@ where
     if let Some(query) = query.clone() {
         req = req.set_json(query);
     }
+    if content_type.is_some() {
+        req = req.insert_header(content_type.clone().unwrap());
+    }
```
chubei, line 71:
```suggestion
    if let Some(content_type) = content_type.cloned() {
        req = req.insert_header(content_type);
    }
```

dozer-api/src/rest/tests/routes.rs
```diff
@@ -63,24 +66,37 @@ where
...
+        count = body.as_u64().unwrap();
+    }
 
     let mut req = actix_web::test::TestRequest::post().uri(&format!("{path}/query"));
     if let Some(query) = query {
         req = req.set_json(query);
     }
+    if content_type.is_some() {
+        req = req.insert_header(content_type.clone().unwrap());
+    }
```
chubei, line 90: Same as above

dozer-api/src/rest/tests/routes.rs
```diff
@@ -63,24 +66,37 @@ where
...
-    let count = body.as_u64().unwrap();
+    let mut count = 0;
+    let mut records: Vec<Value> = vec![];
+    if content_type.is_some() && content_type.clone().unwrap() != ContentType::json() {
+        assert_eq!(res.status(), StatusCode::BAD_REQUEST);
+    } else {
+        assert!(res.status().is_success());
+        let body: Value = actix_web::test::read_body_json(res).await;
+        count = body.as_u64().unwrap();
+    }
```
chubei, line 82:
```suggestion
    res.error_for_status()?;
    let body: Value = actix_web::test::read_body_json(res).await;
    let count = body.as_u64().unwrap();
```

dozer-api/src/rest/tests/routes.rs
```diff
@@ -63,24 +66,37 @@ where
...
-    let body: Value = actix_web::test::read_body_json(res).await;
-    let records = body.as_array().unwrap().to_vec();
-
+    if content_type.is_some() && content_type.clone().unwrap() != ContentType::json() {
+        assert_eq!(res.status(), StatusCode::BAD_REQUEST);
+    } else {
+        assert!(res.status().is_success());
+        let body: Value = actix_web::test::read_body_json(res).await;
+        records = body.as_array().unwrap().to_vec();
+    }
```
chubei, line 99: Same as above

dozer-api/src/rest/tests/routes.rs
```diff
@@ -53,6 +55,7 @@ async fn count_and_query<S, B, E>(
     path: &str,
     service: &S,
     query: Option<Value>,
+    content_type: Option<ContentType>,
 ) -> (u64, Vec<Value>)
```
chubei, line 59:
```suggestion
) -> Result<(u64, Vec<Value>), SomeError>
```

chubei (changes requested):
The test won't fail even if malformed json is not handled correctly

dozer-api/src/rest/tests/routes.rs
aaryaattrey: Used `StatusCode` as `SomeError`. Asserting that the correct `StatusCode` is received.

dozer-api/src/rest/tests/routes.rs
aaryaattrey: `error_for_status` is not present in `actix-web`. Used `res.status().is_success()` to check, and manually return the corresponding status code.

chubei (approved):

PR #1897, opened 2023-08-23
Add merge_group trigger for e2e-test GHA workflow

Jesse-Bakker:


chubei:
We also need to change `schedule` to `pull_request_target` as we did before?

chubei:
But it's added to branch protection. I thought you were going to trigger it and skip it on `pull_request`.

We don't need `pull_request_target` any more.

chubei (approved):

Jesse-Bakker:
I'm not sure what happens if it's required but not triggered. Let's try it without triggering on PR and if it doesn't work, I'll trigger+skip.

PR #1902, opened 2023-08-23
fix: introduce build states and fix inconsistencies

v3g42:
- Broadcasts start, success and failure of build
- Removed recursive watching (This was causing `.dozer` and `target` folders to trigger build)

### DIFF
--- a/dozer-cli/src/live/server.rs
+++ b/dozer-cli/src/live/server.rs
@@ -105,6 +105,7 @@ impl CodeService for LiveServer {
                 .send(Ok(ConnectResponse {
                     live: Some(initial_state),
                     progress: None,
+                    build: None,
                 }))
                 .await
             {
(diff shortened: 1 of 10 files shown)


dozer-cli/src/live/server.rs
```diff
@@ -105,11 +105,13 @@ impl CodeService for LiveServer {
...
                     live: Some(initial_state),
                     progress: None,
+                    build: None,
                 }))
                 .await
             {
                 info!("Error getting initial state");
-                info!("{:?}", e);
+                info!("{}", e.to_string());
+                return {};
```
chubei, line 114: Just `return;`?

dozer-cli/src/live/watcher.rs
```diff
@@ -22,7 +23,13 @@ pub async fn watch(
 
     let watcher = debouncer.watcher();
 
-    watcher.watch(dir.as_path(), RecursiveMode::Recursive)?;
+    watcher.watch(dir.as_path(), RecursiveMode::NonRecursive)?;
```
chubei, line 26: This will work against `dozer init`'s directory structure.

I remember we've already modified `home_dir` to a temporary directory. Maybe we only need to set `cache_dir` to a temporary directory too.

chubei (approved):

chubei:
I'll resolve and merge now to unblock other work.

PR #1906, opened 2023-08-24
fix: Put live cache in temporary directory.

chubei:


v3g42:
@chubei when working with `cargo run`, target folder is getting included and any code change seems to affect rebuilding. 
It is not a production concern but annoying while developing. 
Can we do something about it as well ?

v3g42:
do we have a config-path ? that would be perfect.

chubei:
We use glob patterns for `--config-path` https://github.com/getdozer/dozer/blob/main/dozer-cli/src/config_helper.rs#L21 Looks like `Watcher` doesn't support it https://docs.rs/notify/6.1.1/notify/trait.Watcher.html#tymethod.watch.

This is not trivial. I'll revert the watcher change.

v3g42 (approved):

PR #1907, opened 2023-08-24
chore: Remove `LogOperation::Terminate`

chubei:


Jesse-Bakker (approved):
Seems this operation was never emitted, so this is just a cleanup?

chubei:
Yes it is.

PR #1912, opened 2023-08-24
feat: make probabilistic optimizations optional and tunable in the config

### LINKED ISSUE #1875: Allow switching between accurate vs probabilistic model in pipeline
We're using probabilistic models in the pipeline to improve performance. There are two patterns:

- Using hash of something as the key of a `HashMap`, it appears in [aggregation processor](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/aggregation/processor.rs#L59) and [join processor](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/product/join/operator.rs#L57).
- Using a bloom filter to test for existence. It appears in [set processor](https://github.com/getdozer/dozer/blob/main/dozer-sql/src/pipeline/product/set/set_processor.rs#L20).

We'd like to enable the users to choose between accuracy and performance. This can be a runtime flag the the whole pipeline level. The individual processors will respect this flag and use corresponding implementation.

This requires:

- [ ] Implement "accurate" model for relevant processors.
- [ ] Add the flag to the configuration file.

Implementation note: We should probably clean up join processor before implementing the change. There are quite some duplicated code.

abcpro1:
Resolves #1875. 

Probabilistic optimization sacrifices accuracy in order to reduce memory consumption. In certain parts of the pipeline, a Bloom Filter is used ([set_processor](https://github.com/getdozer/dozer/blob/2e3ba96c3f4bdf9a691747191ab15617564d8ca2/dozer-sql/src/pipeline/product/set/set_processor.rs#L20)), while in other parts, hash tables that store only the hash of the keys instead of the full keys are used ([aggregation_processor](https://github.com/getdozer/dozer/blob/2e3ba96c3f4bdf9a691747191ab15617564d8ca2/dozer-sql/src/pipeline/aggregation/processor.rs#L59) and [join_processor](https://github.com/getdozer/dozer/blob/2e3ba96c3f4bdf9a691747191ab15617564d8ca2/dozer-sql/src/pipeline/product/join/operator.rs#L57-L58)).

This PR makes these optimizations disabled by default and offers user-configurable flags to enable each of these optimizations separately.

This is an example of how to turn-on probabilistic optimizations for each processor in the Dozer configuration.

```yaml
flags:
  enable_probabilistic_optimizations:
    in_sets: true  # enable probabilistic optimizations in set operations (UNION, EXCEPT, INTERSECT); Default: false
    in_joins: true  # enable probabilistic optimizations in JOIN operations; Default: false
    in_aggregations: true  # enable probabilistic optimizations in aggregations (SUM, COUNT, MIN, etc.); Default: false
```

### DIFF
--- a/dozer-cli/src/simple/executor.rs
+++ b/dozer-cli/src/simple/executor.rs
@@ -5,6 +5,7 @@ use dozer_cache::dozer_log::replication::Log;
 use dozer_core::checkpoint::{CheckpointFactory, CheckpointFactoryOptions};
 use dozer_core::processor_record::ProcessorRecordStore;
 use dozer_types::models::api_endpoint::ApiEndpoint;
+use dozer_types::models::flags::Flags;
 use dozer_types::parking_lot::Mutex;
 use tokio::runtime::Runtime;
 
@@ -84,6 +85,7 @@ impl<'a> Executor<'a> {
         runtime: &Arc<Runtime>,
         executor_options: ExecutorOptions,
         shutdown: ShutdownReceiver,
+        flags: Flags,
     ) -> Result<DagExecutor, OrchestrationError> {
         let builder = PipelineBuilder::new(
             self.connections,
@@ -94,6 +96,7 @@ impl<'a> Executor<'a> {
                 .map(|(endpoint, log)| (endpoint.clone(), Some(log.log.clone())))
                 .collect(),
             self.multi_pb.clone(),
+            flags,
         );
 
         let dag = builder.build(runtime, shutdown).await?;
--- a/dozer-sql/src/pipeline/utils/record_hashtable_key.rs
+++ b/dozer-sql/src/pipeline/utils/record_hashtable_key.rs
@@ -1,0 +1,32 @@
+use ahash::AHasher;
+use dozer_types::types::Field;
+use std::hash::{Hash, Hasher};
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum RecordKey {
+    Accurate(Vec<Field>),
+    Hash(u64),
+}
+
+pub fn get_record_hash<'a, I>(fields_iter: I) -> u64
+where
+    I: Iterator<Item = &'a Field>,
+{
+    let mut hasher = AHasher::default();
+    for field in fields_iter {
+        field.hash(&mut hasher);
+    }
+    hasher.finish()
+}
+
+#[test]
+fn test_record_hash() {
+    let record_a = vec![Field::Int(1), Field::String("a".into())];
+    let record_b = vec![Field::Int(2), Field::String("b".into())];
+
+    let hash_a = get_record_hash(record_a.iter());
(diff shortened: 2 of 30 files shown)


dozer-cli/src/simple/executor.rs
```diff
@@ -94,6 +95,7 @@ impl<'a> Executor<'a> {
                 .map(|(endpoint, log)| (endpoint.clone(), Some(log.log.clone())))
                 .collect(),
             self.multi_pb.clone(),
+            Flags::default(),
```
chubei, line 98: I think we'll need to pass flags from config here?

dozer-sql/src/pipeline/utils/record_hashtable_key.rs
```diff
@@ -0,0 +1,32 @@
+use ahash::AHasher;
+use dozer_types::types::Field;
+use std::hash::{Hash, Hasher};
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum RecordKey {
+    Accurate(Vec<Field>),
+    Hash(u64),
```
chubei, line 8: Given that we've already calculated hash in the construction of `RecordKey`, the implementation of `ProbabilisticHashTable` feels redundant.

Can we discard the generality of the code and make `HashTable::K` `RecordKey`?

chubei (commented):
Looking great! Two comments I hope we can discuss.

dozer-cli/src/simple/executor.rs
abcpro1: That's right!

dozer-sql/src/pipeline/utils/record_hashtable_key.rs
abcpro1: You're absolutely right. The abstract HashTable was the first attempt, but when I switched to an abstract Key, it made the special HashTable redundant.

chubei: Sorry but the new usage of `HashMap` seems no different from `ProbabilisticHashTable`, in terms of hash calculating?

We still calculate hashes twice: first to construct `RecordKey`, second to hash `RecordKey` in `HashMap`.

chubei (commented):
https://github.com/getdozer/dozer/pull/1912#discussion_r1305234452

dozer-sql/src/pipeline/utils/record_hashtable_key.rs
abcpro1: Yes, this is how the old implementation used to work. I merely added an accurate implementation besides the old one. I can make a `Hash` trait implementation for `RecordKey` that feeds a precalculated hash to a custom pass-through `Hasher`. What do you think of this idea?

chubei: You're right. Let's merge it as is.

chubei (approved):

PR #1917, opened 2023-08-25
feat: Use custom `CountingBloomFilter` that implements `serde`

chubei:


dozer-sql/src/pipeline/product/set/record_map/bloom.rs
```diff
@@ -0,0 +1,195 @@
...
+            .unwrap_or(0)
+    }
+
+    pub fn clear(&mut self) {
+        self.counters.iter_mut().for_each(|counter| *counter = 0);
+    }
+}
+
+fn optimal_num_counters(num_items: u32, false_positive_rate: f32) -> usize {
+    -(num_items as f32 * false_positive_rate.ln() / (2.0f32.ln().powi(2))).ceil() as usize
```
abcpro1, line 56: You should use `f64`s instead of `f32`s to do the computation in this function. It produces different results using the default parameters `10000000` and `0.01` for `num_items` and `false_positive_rate` accordingly; `95850592` for `f32` and `95850584` for `f64`. The reference you linked uses `f64` in this function, as well as the external `bloom` crate in a function that performs a similar calculation.

abcpro1 (changes requested):
Other than the note I left about `f32`, it is a perfect drop-in replacement. It's also well written.

dozer-sql/src/pipeline/product/set/record_map/bloom.rs
chubei: Good catch!

abcpro1 (approved):
Perfect

Jesse-Bakker:
I am not too familiar with probabilistic data structures, so this may be a stupid question, but what is the difference between a counting bloom filter and hyperloglog?

chubei:
Counting bloom filter is for counting the number of occurrences of every element. You can think this implementation as an approximate of `HashMap<K, u8>`.

It follows the bloom filter that the counting is approximate, in the sense that the tracked count is an upper bound of the actual count.

The value type is `u8` so we can only count to at most 255.

Jesse-Bakker:
Ah, so the difference is in what they count. CBF is count per distinct element (or multiplicity) whereas HLL is number of distinct elements (cardinality of the set). Thanks!

Issue #1926, opened 2023-08-28
`dozer init` can fetch options from a remote server


chubei:
Currently every time we want to update the prompts and templates used in `dozer init`, we have to update dozer code and release new versions of dozer.

It's better if we design a prompt and template configuration file format, and put that configuration on a remote server. This way if we, say, want to add a new connector template, we can just update the configuration file.

PR #1928, opened 2023-08-29
Tempfix: added MongoDB, MySQL and S3 templates to dozer init

supergi0:


chubei:
Why it's a "Tempfix"?

chubei (approved):

supergi0:
In the longer run, we are planning to implement https://github.com/getdozer/dozer/issues/1926, hence thought of this title.

Issue #1915, opened 2023-08-25
Assign a unique ID to metrics on live server between application runs


v3g42:
When you run `dozer live` with different applications today, metrics are intermingled. We have this separation on cloud through deployment_id. 

On a local environment, I suggest adding that tag in https://github.com/getdozer/dozer/blob/main/dozer-cli/src/live/progress.rs

Expected behavior
- Repeat restarts should start counters from 0
- Run multiple applications one after another, metrics should stay isolated on the UI.

chubei:
How about just delete data of older runs?

chubei:
Did we solve the metrics problem @v3g42?

v3g42:
Yes.

PR #1943, opened 2023-08-30
feat: Label metrics from live run with "live_run_id"

### LINKED ISSUE #1940: After Start/Stop in Dozer Live data keeps accumulating in the progress
If I start / stop / start dozer live the counts in the progress bar keeps increasing.
After a stop and a restart it should restart from 0

chubei:
Closes #1940 

Also the backend part of https://github.com/getdozer/dozer/issues/1915


Jesse-Bakker:
Can we make a `metrics` `Recorder` that drives the progress bars instead of having to update the progress bars from components like `cache_builder`?

chubei:
I don't want to keep track of a `HashMap<String, ProgressBar>` because that'll requrie interior mutability. Any way around this?

chubei:
Oh I misunderstood. You meant to drive progress bar from metrics. I believe that's out of the scope of this PR.

dozer-cli/src/live/progress.rs
```diff
@@ -61,3 +64,19 @@ pub async fn progress_stream(
         retry_interval.tick().await;
     }
 }
+
+fn labels_match(
+    prom_labels: &prometheus_parse::Labels,
+    dozer_labels: &dozer_tracing::Labels,
+) -> bool {
+    for (key, value) in dozer_labels.iter() {
```
Jesse-Bakker, line 72:
```rust
dozer_labels.iter().all(|(key, value)| prom_labels.get(key) == Some(value))
```

Jesse-Bakker (approved):
Small style nit, otherwise LGTM!

dozer-cli/src/live/progress.rs
chubei: Much nicer!

snork-alt:
@chubei Why does this require doc update ?

chubei:
Sorry bogus auto labelling. Removing now.

PR #1945, opened 2023-08-30
Write dozer.lock

Jesse-Bakker:
Currently, this always writes the lock file to the current working directory.
Ideally we would follow the `dozer-config.yaml` around, but that is currently
impractical because of the implementation of the `-c` flag, combining `*.sql` and
`dozer-config.*`.

### DIFF
--- a/dozer-cli/src/simple/build/contract/mod.rs
+++ b/dozer-cli/src/simple/build/contract/mod.rs
@@ -4,7 +4,7 @@ use std::{
     path::Path,
 };
 
-use dozer_cache::dozer_log::{home_dir::BuildPath, schemas::EndpointSchema};
+use dozer_cache::dozer_log::schemas::EndpointSchema;
 use dozer_core::{
     dag_schemas::DagSchemas,
     daggy::{self, NodeIndex},
@@ -162,13 +162,13 @@ impl Contract {
         })
     }
 
-    pub fn serialize(&self, build_path: &BuildPath) -> Result<(), BuildError> {
-        serde_json_to_path(&build_path.dag_path, &self)?;
+    pub fn serialize(&self, path: &Path) -> Result<(), BuildError> {
+        serde_json_to_path(path, &self)?;
         Ok(())
     }
 
-    pub fn deserialize(build_path: &BuildPath) -> Result<Self, BuildError> {
-        serde_json_from_path(&build_path.dag_path)
+    pub fn deserialize(path: &Path) -> Result<Self, BuildError> {
+        serde_json_from_path(path)
     }
 }
 
@@ -222,6 +222,7 @@ fn serde_json_to_path(path: impl AsRef<Path>, value: &impl Serialize) -> Result<
     let file = OpenOptions::new()
         .create(true)
         .write(true)
+        .truncate(true)
         .open(path.as_ref())
         .map_err(|e| BuildError::FileSystem(path.as_ref().into(), e))?;
     serde_json::to_writer_pretty(file, value).map_err(BuildError::SerdeJson)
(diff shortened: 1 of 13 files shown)


chubei (changes requested):
With this implementation, I think we can get rid of `BuildPath`.

It was designed to contain all information of a specific dozer execution, so user can restart dozer and continue from checkpoint.
Now that the contract file is moved out of it, there's no way to use data left in there anymore, so it makes more sense to just remove `BuildPath` and put everything that was previously in `BuildPath` to the `HomeDir`.

dozer-cli/src/simple/build/contract/mod.rs
```diff
@@ -219,12 +219,15 @@ fn collect_ancestor_sources_recursive(
 }
 
 fn serde_json_to_path(path: impl AsRef<Path>, value: &impl Serialize) -> Result<(), BuildError> {
-    let file = OpenOptions::new()
+    let mut file = OpenOptions::new()
```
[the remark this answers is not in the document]
Jesse-Bakker: The truncate is needed, I think, but yeah, the rest is debug code

Jesse-Bakker:
If the contract is consistent with the endpoints, we can still re-use the build, the same way as when the contract was written to the build dir, I think.

dozer-cli/src/simple/orchestrator.rs
```diff
@@ -270,7 +294,9 @@ impl SimpleOrchestrator {
         force: bool,
         shutdown: ShutdownReceiver,
     ) -> Result<(), OrchestrationError> {
-        let home_dir = HomeDir::new(self.config.home_dir.clone(), self.config.cache_dir.clone());
+        let home_dir = self.base_directory.join(&self.config.home_dir);
+        let cache_dir = self.base_directory.join(&self.config.cache_dir);
```
chubei, line 298: Use `self.home_dir` and such

dozer-cli/src/simple/orchestrator.rs
```diff
@@ -327,24 +353,33 @@ impl SimpleOrchestrator {
...
+        contract.serialize(contract_path.as_std_path())?;
 
         Ok(())
     }
 
     // Cleaning the entire folder as there will be inconsistencies
     // between pipeline, cache and generated proto files.
     pub fn clean(&mut self) -> Result<(), OrchestrationError> {
-        let cache_dir = PathBuf::from(self.config.cache_dir.clone());
+        let cache_dir = PathBuf::from(self.base_directory.join(&self.config.cache_dir));
```
chubei, line 376: Same as above

dozer-cli/src/simple/orchestrator.rs
```diff
@@ -327,24 +353,33 @@ impl SimpleOrchestrator {
...
         if cache_dir.exists() {
             fs::remove_dir_all(&cache_dir)
                 .map_err(|e| ExecutionError::FileSystemError(cache_dir, e))?;
         };
 
-        let home_dir = PathBuf::from(self.config.home_dir.clone());
+        let home_dir = PathBuf::from(self.base_directory.join(&self.config.home_dir));
```
chubei, line 382: Same as above

chubei (approved):
Some minor comments

chubei:
> the same way as when the contract was written to the build dir

I didn't make myself clear.

Before this PR, it's possible to implement such a command:

`dozer run app --build 1`

where we compare the "current" contract with the contract stored in build `v0001`. If they are compatible, we start from checkpoint. If not, we report error.

Now that contract does not exist in the build directory anymore, such command is not possible. And this beats the purpose of the existence of build directories.

That's why I said we can get rid of `BuildPath` now. It's not useful anymore.

Jesse-Bakker:
Ah, as there is only ever one live build, the home dir in practice becomes the build directory. Luckily the contents of the home dir are an implementation detail, so I will defer this to after versioning

PR #1951, opened 2023-08-31
feat: Add --locked flag to build and run

Jesse-Bakker:
Depends on #1945

### DIFF
--- a/dozer-cli/src/simple/build/contract/mod.rs
+++ b/dozer-cli/src/simple/build/contract/mod.rs
@@ -10,6 +10,7 @@ use dozer_core::{
     daggy::{self, NodeIndex},
     node::PortHandle,
     petgraph::{
+        algo::is_isomorphic_matching,
         visit::{EdgeRef, IntoEdgesDirected, IntoNodeReferences},
         Direction,
     },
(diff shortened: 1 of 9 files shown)


dozer-cli/src/simple/build/contract/mod.rs
```diff
@@ -239,4 +240,17 @@ where
     serde_json::from_reader(file).map_err(BuildError::FailedToLoadExistingContract)
 }
 
+impl PartialEq<Contract> for Contract {
+    fn eq(&self, other: &Contract) -> bool {
+        self.version == other.version
```
chubei, line 245: This worries me a little that when we add new fields to `Contract` we'll forget to update `PartialEq` implementation.

Maybe we can do
```rust
let Contract {
  version: self_version,
  pipeline: self_pipeline,
  endpoint: self_endpoints,
} = self;
let Contract {
  version: other_version,
  pipeline: other_pipeline,
  endpoint: other_endpoints,
} = other;
```

And then compare them.

chubei (approved):
`is_isomorphic_matching` is handy!

dozer-cli/src/simple/build/contract/mod.rs
Jesse-Bakker: Or maybe wrap pipeline in a newtype and impl partialeq on that and derive it on Contract?

chubei: Even better

chubei: The new type is called `PipelineContract` by the way. It's already a alias type.

Issue #1916, opened 2023-08-25
Handle network errors in connectors


chubei:
- [ ] postgres
- [ ] mysql
- [ ] snowflake
- [ ] kafka
- [ ] object store

PR #1956, opened 2023-08-31
fix: resume connectors on network errors

abcpro1:
This PR includes MySQL, Postgres, and Kafka connectors, so far.

Covers three connectors from #1916.

### DIFF
--- a/dozer-ingestion/src/connectors/kafka/stream_consumer_basic.rs
+++ b/dozer-ingestion/src/connectors/kafka/stream_consumer_basic.rs
@@ -13,7 +13,6 @@ use dozer_types::serde::{Deserialize, Serialize};
 use dozer_types::serde_json;
 use dozer_types::serde_json::Value;
 use dozer_types::types::{Field, Operation, Record};
-use rdkafka::consumer::BaseConsumer;
 
 use crate::connectors::kafka::no_schema_registry_basic::NoSchemaRegistryBasic;
 use crate::connectors::kafka::schema_registry_basic::SchemaRegistryBasic;
@@ -21,7 +20,9 @@ use tonic::async_trait;
 
 use crate::connectors::TableInfo;
 use crate::errors::KafkaStreamError::PollingError;
-use rdkafka::Message;
+use rdkafka::{ClientConfig, Message};
+
+use super::stream_consumer_helper::{is_network_failure, OffsetsMap, StreamConsumerHelper};
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(crate = "dozer_types::serde")]
@@ -79,17 +80,17 @@ pub struct Payload {
 #[derive(Default)]
 pub struct StreamConsumerBasic {}
 
-impl StreamConsumerBasic {}
-
 #[async_trait]
 impl StreamConsumer for StreamConsumerBasic {
     async fn run(
         &self,
-        con: BaseConsumer,
+        client_config: ClientConfig,
         ingestor: &Ingestor,
         tables: Vec<TableInfo>,
         schema_registry_url: &Option<String>,
     ) -> Result<(), ConnectorError> {
+        let topics: Vec<String> = tables.iter().map(|t| t.name.clone()).collect();
+
         let mut schemas = HashMap::new();
         for (table_index, table) in tables.into_iter().enumerate() {
             let schema = if let Some(url) = schema_registry_url {
@@ -101,10 +102,19 @@ impl StreamConsumer for StreamConsumerBasic {
             schemas.insert(table.name.clone(), (table_index, schema));
         }
 
+        let topics: Vec<&str> = topics.iter().map(|t| t.as_str()).collect();
+        let mut con = StreamConsumerHelper::start(&client_config, &topics).await?;
+
+        let mut offsets = OffsetsMap::new();
         let mut counter = 0;
         loop {
             if let Some(result) = con.poll(None) {
+                if matches!(result.as_ref(), Err(err) if is_network_failure(err)) {
+                    con = StreamConsumerHelper::resume(&client_config, &topics, &offsets).await?;
+                    continue;
+                }
                 let m = result.map_err(|e| KafkaStreamError(PollingError(e)))?;
+                StreamConsumerHelper::update_offsets(&mut offsets, &m);
                 match schemas.get(m.topic()) {
                     None => return Err(ConnectorError::KafkaError(TopicNotDefined)),
                     Some((table_index, (schema, fields_map))) => {
--- a/dozer-ingestion/src/connectors/kafka/stream_consumer_helper.rs
+++ b/dozer-ingestion/src/connectors/kafka/stream_consumer_helper.rs
@@ -1,0 +1,102 @@
+use crate::errors::ConnectorError;
+use crate::errors::KafkaError::KafkaConnectionError;
+use rdkafka::{
+    consumer::{BaseConsumer, Consumer},
+    message::BorrowedMessage,
+    util::Timeout,
+    ClientConfig, Message, Offset,
+};
+use std::collections::HashMap;
+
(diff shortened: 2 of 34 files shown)


dozer-ingestion/src/connectors/postgres/connection/client.rs
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            match ready!(this.inner.as_mut().poll_next(cx)) {
+                Some(Ok(row)) => {
+                    this.cursor_position += 1;
+                    return Poll::Ready(Some(Ok(row)));
+                }
+                Some(Err(err)) => {
+                    if is_network_failure(&err) {
+                        this.pending_resume = Some(this.resume());
```
chubei, line 236: Once we resume, this is a branch new stream and will generate all rows again?

I see `cursor_position` which looks like for dedup, but it's not used?

dozer-ingestion/src/utils.rs
```diff
@@ -0,0 +1,21 @@
+#[macro_export]
+macro_rules! retry_on_network_failure {
```
chubei, line 2: While I appreciate the effort of trying to reuse this piece of code everywhere, I'm a little uneasy that such a complicated macro is used extensively. It took me half an hour to understand what it does!

Per my understanding, this macro only has to be a macro instead of a function because the usage in `exec_iter_impl`, where we pass `continue 'main` as the `reconnect`, which is not something a function can do.

If this understanding is correct, how about we change the macro to a function and write a similar macro in `exec_iter_impl` just for being used in that function?

This macro can be changed to a function like:

```rust
/// Connection network failure should be handled in `reconnect`
async fn retry_on_network_failure<
    Operation,
    OperationFuture,
    ErrorPredicate,
    Reconnect,
    ReconnectFuture,
    Output,
    Error
>(
    description: &'static str,
    mut operation: Operation,
    network_error_predicate: ErrorPredicate,
    mut reconnect: Option<Reconnect>
) -> Result<Output, Error> 
where
    Operation: FnMut() -> OperationFuture,
    OperationFuture: Future<Output = Result<Output, Error>>,
    ErrorPredicate: FnMut(&Error) -> bool,
    Reconnect: FnMut() -> ReconnectFuture,
    ReconnectFuture: Future<Output = Result<(), Error>>,
{
    ...
}
```

dozer-ingestion/src/utils.rs
```diff
@@ -0,0 +1,21 @@
...
+        loop {
+            match $operation {
+                ok @ Ok(_) => break ok,
+                Err(err) => {
+                    if ($network_error_predicate)(&err) {
+                        dozer_types::log::error!(
+                            "network error during {}: {err:?}. retrying...",
+                            $description
+                        );
+                        $($reconnect)?
```
chubei, line 13: We may want to sleep for a little while when network failure happens.

chubei (commented):
This is great!

I left some comments. Let's discuss.

dozer-ingestion/src/connectors/mysql/connection.rs
```diff
@@ -0,0 +1,227 @@
...
+            None => Poll::Ready(None),
+        }
+    }
+}
+
+impl Drop for QueryResult {
+    fn drop(&mut self) {
+        self.receiver.close()
+    }
+}
```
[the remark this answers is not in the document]
abcpro1: Right.

dozer-ingestion/src/connectors/postgres/connection/client.rs
abcpro1: It's used here
https://github.com/getdozer/dozer/blob/07094fd6e9cf7749d00937bf210ff0604e6eaeb8/dozer-ingestion/src/connectors/postgres/connection/client.rs#L203
https://github.com/getdozer/dozer/blob/07094fd6e9cf7749d00937bf210ff0604e6eaeb8/dozer-ingestion/src/connectors/postgres/connection/client.rs#L194
The query isn't restarted as is, the offset `cursor_position` is appended to the SQL query.

chubei: Oh cool. Missed that.

dozer-ingestion/src/utils.rs
abcpro1: Managing the lifetime of `self` in futures is complicated, and with `loop` it gets really frustrating.

chubei: So this function signature doesn't work? I was thinking we can get away from that with all the `Operation`, `OperationFuture`, ... generic types.

abcpro1: Having trouble with `captured variable cannot escape FnMut closure body` with `self.reconnect()` in 
```rust
retry_on_network_failure(
            "prepare",
            || self.inner.prepare(query),
            is_network_failure,
            Some(|| self.reconnect())
```

chubei: Let me play with it for a little. If it's a deep rabbit hole we'll just leave it there.

chubei: OK this is indeed deep. https://smallcultfollowing.com/babysteps/blog/2023/03/29/thoughts-on-async-closures/. I give up.

abcpro1: I sure learned a few lessons this week about rust lifetime management, self-referential structs, and loops. I spent a lot of time solving and dodging these lifetime issues.

chubei: That's only natural for every Rust programmer

dozer-ingestion/src/connectors/kafka/stream_consumer_helper.rs
```diff
@@ -0,0 +1,102 @@
...
+        MessageConsumption(error_code) => error_code,
+        MessageProduction(error_code) => error_code,
+        MetadataFetch(error_code) => error_code,
+        OffsetFetch(error_code) => error_code,
+        Rebalance(error_code) => error_code,
+        SetPartitionOffset(error_code) => error_code,
+        StoreOffset(error_code) => error_code,
+        MockCluster(error_code) => error_code,
+        Transaction(rdkafka_err) => return rdkafka_err.is_retriable(),
+        _ => todo!(),
```
chubei, line 86: Maybe emit a warning and return false?

dozer-ingestion/src/connectors/kafka/stream_consumer_basic.rs
```diff
@@ -79,17 +80,17 @@ pub struct Payload {
...
 impl StreamConsumer for StreamConsumerBasic {
     async fn run(
         &self,
-        con: BaseConsumer,
+        client_config: ClientConfig,
         ingestor: &Ingestor,
         tables: Vec<TableInfo>,
         schema_registry_url: &Option<String>,
     ) -> Result<(), ConnectorError> {
+        let topics: Vec<String> = tables.iter().map(|t| t.name.clone()).collect();
```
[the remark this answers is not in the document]
chubei: Sorry bad sight

dozer-ingestion/src/connectors/kafka/stream_consumer_helper.rs
abcpro1: Good catch! That should just return false without a warning actually because there are many errors that are not network errors.

chubei: Yes I've made that change. Please give me push access to your repo so I can add it to this PR.

chubei (commented):
LGTM now. If you find https://github.com/getdozer/dozer/pull/1956/commits/56d1b767caa035f222a860ca04d0e6228d1057d9 OK, please resolve the conversations to merge.

chubei (approved):

chubei (approved):

PR #1971, opened 2023-09-04
chore: Remove duplicated code in `JoinOperator`

chubei:
This PR only removes duplicated code. No behaviour is changed.


dozer-sql/src/pipeline/product/join/operator/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        let mut output_records = vec![];
+        for matching_record in table_to_match.get_matching_records(join_key, false) {
+            let join_record = join_records(matching_record);
+
+            if need_to_act_on_default_record {
+                let default_join_record = create_join_records_fn(
+                    matching_record,
+                    matching_record_branch,
+                )(table_of_record.default_record());
```
abcpro1, line 143: Maybe you don't need `matching_record_branch`. It could then be moved outside the loop.
```rust
            let default_join_records = create_join_records_fn(
                table_of_record.default_record(),
                record_branch,
            );
            for ... {
                ...
                if need_to_act_on_default_record {
                    let default_join_record = default_join_records(matching_record);
```

abcpro1 (commented):
This looks like an excellent refactoring. It inspires me to write cleaner code.
I have left a comment about the last commit.

dozer-sql/src/pipeline/product/join/operator/mod.rs
chubei: Wonderful point.

mediuminvader (approved):

abcpro1 (approved):

PR #1973, opened 2023-09-05
fix: Use span for orchestrator error

karolisg:


karolisg:
> Correct me if I'm wrong. Is this actually the following one-line change?
> 
> ```rust
> .map_err(|e| {
>         let _span = error_span!("OrchestrationError", error = %e);
> 
>         e
>     })
> ```

Yes, you're right

PR #1975, opened 2023-09-05
check for empty endpoint array

aaryaattrey:


dozer-api/src/grpc/shared_impl/mod.rs
```diff
@@ -97,6 +97,10 @@ pub fn on_event<T: Send + 'static>(
         ));
     }
 
+    if endpoints.is_empty() {
+        return Err(Status::invalid_argument("empty-endpoints-array"));
```
chubei, line 101: Don't add `-`s.

chubei (changes requested):

dozer-api/src/grpc/shared_impl/mod.rs
aaryaattrey: Spaces are not handled properly in the output. It is shown like this in output: `empty%20endpoints%20array`. Is that fine?

chubei: I believe it's frontend problem that the message is not properly displayed. Try postman?

aaryaattrey: Right. Its a frontend issue.

PR #1982, opened 2023-09-05
chore: Remove source state merging from `ReceiverLoop` to `EpochManager`

chubei:
I believe the merging was in `ReceiverLoop` because the `SourceStates` was introduced before `EpochManager`, and every source sent commit messages independently. In that case, `ReceiverLoop`s (processors and sinks) are the only place to merge source states.

Now that we have `EpochManager`, it makes more sense to merge once only in it.


Jesse-Bakker (approved):
Implementation LGTM

What is `SourceStates` for, e.g. what does the `OpIdentifier` signify? Do we maybe want to rename the type or add a docstring to make it clear?

chubei:
Right, I'll add some preliminary doc string.

PR #1991, opened 2023-09-06
chore: read env DOZER_MASTER_SECRET for api security

duonganhthu43:
- [x] preserve Api Security config in live exeperience
- [x]  add `enable_api_security` indicatior in live state to know wherether API explorer should fetch token  before hand
- [x]  Include ApiExplorerService for both `live` and `cloud` experience => this one providing endpoint in order to fetch the short live token on Api Explorer
- [x] using DOZER_MASTER_SECRET variable as default api security config on cloud

### DIFF
--- a/dozer-api/src/grpc/client_server.rs
+++ b/dozer-api/src/grpc/client_server.rs
@@ -121,7 +121,8 @@ impl ApiServer {
         let health_service = web_config.enable(health_service);
 
         // Auth middleware.
-        let auth_middleware = AuthMiddlewareLayer::new(self.security.clone());
+        let security = get_api_security(self.security.to_owned());
+        let auth_middleware = AuthMiddlewareLayer::new(security.clone());
 
         // Authenticated services.
         let common_service = auth_middleware.layer(common_service);
--- a/dozer-api/src/grpc/client_server.rs
+++ b/dozer-api/src/grpc/client_server.rs
@@ -1,5 +1,6 @@
 use super::metric_middleware::MetricMiddlewareLayer;
 use super::{auth_middleware::AuthMiddlewareLayer, common::CommonService, typed::TypedService};
+use crate::api_helper::get_api_security;
 use crate::errors::ApiInitError;
 use crate::grpc::auth::AuthService;
 use crate::grpc::health::HealthService;
@@ -56,11 +57,7 @@ impl ApiServer {
             builder = builder.register_encoded_file_descriptor_set(descriptor_bytes);
         }
         let inflection_service = builder.build().map_err(GrpcError::ServerReflectionError)?;
-        let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
-        let security = self
-            .security
-            .clone()
-            .or_else(|| dozer_master_secret.map(ApiSecurity::Jwt));
+        let security = get_api_security(self.security.to_owned());
 
         // Service handling dynamic gRPC requests.
         let typed_service = if self.flags.dynamic {
@@ -139,15 +136,10 @@ impl ApiServer {
         let health_service = auth_middleware.layer(health_service);
 
         let mut auth_service = None;
-        let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
-        let security = if self.security.is_none() && dozer_master_secret.is_some() {
-            Some(ApiSecurity::Jwt(dozer_master_secret.unwrap()))
-        } else {
-            self.security.clone()
-        };
+        let security = get_api_security(self.security.to_owned());
         if security.is_some() {
             let service = web_config.enable(AuthGrpcServiceServer::new(AuthService::new(
-                self.security.clone(),
+                security.to_owned(),
             )));
             auth_service = Some(auth_middleware.layer(service));
         }
@@ -180,11 +172,9 @@ impl ApiServer {
         let addr = format!("{}:{}", self.host, self.port);
         info!(
             "Starting gRPC server on {addr} with security: {}",
-            self.security
-                .as_ref()
-                .map_or("None".to_string(), |s| match s {
-                    ApiSecurity::Jwt(_) => "JWT".to_string(),
-                })
+            security.as_ref().map_or("None".to_string(), |s| match s {
+                ApiSecurity::Jwt(_) => "JWT".to_string(),
+            })
         );
         let addr = addr
             .parse()
--- a/dozer-api/src/grpc/client_server.rs
+++ b/dozer-api/src/grpc/client_server.rs
@@ -57,11 +57,11 @@ impl ApiServer {
         }
         let inflection_service = builder.build().map_err(GrpcError::ServerReflectionError)?;
         let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
-        let security = if self.security.is_none() && dozer_master_secret.is_some() {
-            Some(ApiSecurity::Jwt(dozer_master_secret.unwrap()))
-        } else {
-            self.security.clone()
-        };
+        let security = self
+            .security
+            .clone()
+            .or_else(|| dozer_master_secret.map(ApiSecurity::Jwt));
+
         // Service handling dynamic gRPC requests.
         let typed_service = if self.flags.dynamic {
             Some(TypedService::new(
--- a/dozer-api/src/grpc/client_server.rs
+++ b/dozer-api/src/grpc/client_server.rs
@@ -56,13 +56,18 @@ impl ApiServer {
             builder = builder.register_encoded_file_descriptor_set(descriptor_bytes);
         }
         let inflection_service = builder.build().map_err(GrpcError::ServerReflectionError)?;
-
+        let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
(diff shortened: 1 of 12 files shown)


dozer-api/src/grpc/client_server.rs
```diff
@@ -56,13 +56,18 @@ impl ApiServer {
             builder = builder.register_encoded_file_descriptor_set(descriptor_bytes);
         }
         let inflection_service = builder.build().map_err(GrpcError::ServerReflectionError)?;
+        let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
+        let security = self
+            .security
+            .clone()
+            .or_else(|| dozer_master_secret.map(ApiSecurity::Jwt));
```
chubei, line 63: Learnt from @karolisg, environment variable should override configuration.

dozer-api/src/grpc/client_server.rs
```diff
@@ -134,7 +139,13 @@ impl ApiServer {
         let health_service = auth_middleware.layer(health_service);
 
         let mut auth_service = None;
-        if self.security.is_some() {
+        let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
+        let security = if self.security.is_none() && dozer_master_secret.is_some() {
+            Some(ApiSecurity::Jwt(dozer_master_secret.unwrap()))
+        } else {
+            self.security.clone()
+        };
```
chubei, line 147: Looks like duplicated code with the lines above

dozer-api/src/rest/mod.rs
```diff
@@ -165,18 +165,21 @@ impl ApiServer {
         shutdown: impl Future<Output = ()> + Send + 'static,
         labels: LabelsAndProgress,
     ) -> Result<Server, ApiInitError> {
+        let dozer_master_secret = std::env::var("DOZER_MASTER_SECRET").ok();
+        let security = match (dozer_master_secret, self.security) {
+            (Some(master_secret), None) => Some(ApiSecurity::Jwt(master_secret)),
+            (_, security) => security,
+        };
```
chubei, line 172: This logic is different from what's used in `client_server.rs`. Here environment variable overrides configuration.

dozer-cli/src/live/state.rs
```diff
@@ -167,6 +167,13 @@ impl LiveState {
                 app_name: dozer.dozer.config.app_name.clone(),
                 connections,
                 endpoints,
+                enable_api_security: dozer
+                    .dozer
+                    .config
+                    .api
+                    .clone()
```
chubei, line 174: No need to `clone`, `as_ref` should be enough. And this logic is different from above in that it doesn't use env var.

chubei (changes requested):

chubei:
I don't get the purpose of implementing `DOZER_MASTER_TOKEN` env var.

dozer-api/src/grpc/client_server.rs
duonganhthu43: hmm, i think should not in this case, we should prior the secret user input in `dozer-config.yaml` right, since they have no idea about this variable. We only take this value if user does not specify secret in api security config

dozer-api/src/rest/mod.rs
duonganhthu43: hmm, i think the code above, logic will be: if api security is empty, we will take value from variable DOZER_MASTER_SECRET => which mean we prior config over variable

dozer-api/src/grpc/client_server.rs
chubei: From dozer core's point of view, env var and config are both user input.

It's the cloud orchestrator manipulating user input, and it should (and it already does) ensure that it respects value in config.

dozer-api/src/grpc/client_server.rs
chubei: Yes. I meant the code is duplicated. We can create a function to avoid repeating ourselves.

dozer-api/src/rest/mod.rs
chubei: Oh sorry my bad. Again we should create a function to avoid repeating this bit of logic

dozer-api/src/grpc/client_server.rs
duonganhthu43: sure, let me update with that rule: variable should override configuration

dozer-types/protos/api_explorer.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+package dozer.api_explorer;
+service ApiExplorerService {
+    rpc GetApiToken(GetApiTokenRequest) returns (GetApiTokenResponse);
+}
+
+message GetApiTokenRequest {
+  // required if in cloud context.
+  optional string app_id = 1;
```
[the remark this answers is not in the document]
duonganhthu43: right now, in Cloud context, the deployment number is not really helpful in this case, since we have Variable per application, not per deployment. So app_id is enough for fetching secret to generate token. I think we have to revise this with the case multiple APIs is running at same time. Should secret token is different per API ?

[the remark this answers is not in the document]
chubei: Hmm, I guess user has to live with the fact that api secret cannot be changed.

chubei (dismissed):

dozer-types/protos/api_explorer.proto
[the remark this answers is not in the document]
chubei: I have one more question, what would happen if the user uses `{{}}` syntax and pass api secret as a secret?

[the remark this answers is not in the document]
duonganhthu43: right, i still not consider this case. It gonna conflict with the rule we have previously: Prior Env over config
For instance, we input `!JWT {{VARIABLE_SECRET}} `
=> sending to cloud and deploy => env DOZER_MASTER_SECRECT = {{VARIABLE_SECRET}}
=> when the `dozer` binaray is called inside pod, `!JWT {{VARIABLE_SECRET}} ` will be binding with actuall value => so the Api Config will have the correct value, not the DOZER_MASTER_SECRECT

=> Maybe for this case, on `dozer` repo logic, we should prior the config over variable ? What do you think about this @chubei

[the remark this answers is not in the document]
chubei: Agreed, prior the config over variable will solve this problem.

Another problem remains, the `DOZER_MASTER_SECRECT` being the literal `{{VARIABLE_SECRET}}` will prevent the `ApiExplorerService` from working.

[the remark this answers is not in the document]
chubei: How about giving instructions to the user for generating token and guide them to put into the API explorer?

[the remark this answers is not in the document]
duonganhthu43: > How about giving instructions to the user for generating token and guide them to put into the API explorer?

hmm, that's not a really convenient experience on API Explore
I was thinking a way like this: 
on `dozer` binary, when running, we add extra step logic =>  if api config jwt having value, we set env DOZER_MASTER_SECRECT = jwt secret value
=> it should work right ?, so next time we query DOZER_MASTER_SECRECT from ApiExplorerService, correct value should be there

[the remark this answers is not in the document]
chubei: If we're going to make the secret plain text, we can just read it from configuration and save it to k8s, without going through environment variable. And I'm not sure if we can let a container pass information to k8s through environment variables.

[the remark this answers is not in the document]
duonganhthu43: oh, then a more simple approach, everything stay the same
we have a list of secrets already
just try bind DOZER_MASTER_SECRET with `handlebars` before send it to `deployer` along with secrets list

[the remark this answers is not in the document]
duonganhthu43: yup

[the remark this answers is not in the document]
duonganhthu43: include this logic on cloud [PR](https://github.com/getdozer/dozer-cloud/pull/554/files)
can you check and approve it also @chubei  ? these PRs should go together

dozer-cli/src/live/state.rs
duonganhthu43: updated using as_ref

duonganhthu43: not really get it ??

duonganhthu43: ah, i see. Since `live state` is for local running. User should only have api security from `dozer-config.yaml`.  Hardly the `DOZER_MASTER_TOKEN ` is manually set by user - only if they know about it - Only Cloud should use it when deploy. Anyway, i update it to ensure the synchronize logic

dozer-cli/src/live/state.rs
```diff
@@ -164,10 +165,23 @@ impl LiveState {
                 .iter()
                 .map(|c| c.name.clone())
                 .collect();
+
+            let enable_api_security = std::env::var("DOZER_MASTER_SECRET")
```
chubei, line 169: Should we reuse `get_api_security`?

chubei (approved):

PR #1992, opened 2023-09-06
chore: use struct to send query in grpc

v3g42:


v3g42:
This is no longer relevant.

PR #2005, opened 2023-09-07
change on_event to OnEvent

aaryaattrey:
Use method name OnEvent for both common and typed service.


chubei:
Does live still work?

aaryaattrey:
Yes it works.

PR #2007, opened 2023-09-07
fix: mysql select query offset syntax is wrong

abcpro1:
Reference: https://dev.mysql.com/doc/refman/8.0/en/select.html#:~:text=To%20retrieve%20all,%2C18446744073709551615%3B


abcpro1:
It's `u64::MAX`.

```bash
$ python -c 'print(2**64 - 1)'
18446744073709551615
```

PR #2015, opened 2023-09-11
chore: remove old methods

v3g42:


dozer-types/protos/cloud.proto
```diff
@@ -5,21 +5,12 @@ import "google/protobuf/timestamp.proto";
 
 service DozerCloud {
   rpc create_application(CreateAppRequest) returns (AppResponse);
```
chubei, line 7: Do we want to also remove `create_application, `update_application` and `get_application`?

dozer-types/protos/cloud.proto
```diff
@@ -52,93 +40,103 @@ service DozerCloud {
...
+  bool healthy = 1;
+}
+message GetResourcesRequest {
   string app_id = 1;
-  optional uint32 limit = 2;
-  optional uint32 offset = 3;
+  // if not specified, will get latest version
+  optional uint32 version = 2;
+}
+message ResourcesResponse {
```
chubei, line 126: Looks like renamed from `DeploymentStatusResponse`. Why remove `DeploymentStatusResponse::errors`?

dozer-types/protos/cloud.proto
```diff
@@ -52,93 +40,103 @@ service DozerCloud {
...
+  repeated DeploymentResource resources = 1;
+  map<uint32, uint32> versions = 3;
+  uint32 current_version = 4;
 }
 
+message DeploymentResource {
```
chubei, line 132: Looks like renamed from `DeploymentHealth`. Why remove `DeploymentHealth::status`?

dozer-types/protos/cloud.proto
v3g42: Lets do this in another PR. As @karolisg is still figuring out best way to deal with secrets. Currently using secrets doesn't work as application is not created yet.

dozer-types/protos/cloud.proto
chubei: I get the state reduction part. It's just that we're not getting errors anymore.

dozer-types/protos/cloud.proto
v3g42: @Jesse-Bakker's recent PR introduces Job type for Applications. This complicates things a little bit as payload for `Job` and `Deployment` are a little different through APIs. So we can delegate status calculation to the frontend.

chubei: I remember `status` was from k8s resource status. It's not returned to frontend?

dozer-types/protos/cloud.proto
v3g42: These errors were not really errors. They were just returning the last message in each pod which was rather confusing. 
Instead we should have an API that either calls`get_logs` or `get_errors` and create a separate view for it.

chubei: Makes sense.

dozer-types/protos/cloud.proto
v3g42: not in the same form. It was being transformed.

chubei: Yes, I meant that frontend now doesn't have the information that was used to calculate `status`. It only has `desired`, `available`, etc.

v3g42: https://github.com/getdozer/dozer-frontend/pull/71/files#diff-bd9ee897c857753c2ace2bdb9e4d7aefcaed16bd93c86e9bd6e8e031a76d1e98
This wasnt comprehensive before.

chubei: I'm not sure about what you mean.

As far as I can see, new `status` is calculated based on `desired` and `available`: https://github.com/getdozer/dozer-frontend/pull/71/files#diff-bd9ee897c857753c2ace2bdb9e4d7aefcaed16bd93c86e9bd6e8e031a76d1e98R55
While old `status` is calculated based on pod conditions: https://github.com/getdozer/dozer-cloud/pull/570/files#diff-2579a5f330b2b0551e29104778b207398caa2493197ee3d5c5ca6b0067917daaL11

v3g42: Which is exactly what i was trying to avoid to be done on the backend.

chubei (approved):

PR #2019, opened 2023-09-12
chore: Enable `tokio-console` traces

chubei:


Jesse-Bakker:
I think I'd rather put tokio tracing and console-subscriber behind a cargo feature. That way it's still easy to reach for `tokio-console` when you need it, but we don't incur its overhead (not sure how big) by default. Furthermore, when someone deploys Dozer, it's another port they'd have to lock down.

chubei:
Makes sense. Let me change that.

chubei:
@Jesse-Bakker  I'm not sure what to do with the rustflags in `.cargo/config.toml`. It can't be controlled by cargo features right?

Jesse-Bakker:
The only way to control a cfg with a feature would be with a `build.rs`, which seems like overkill. I think all tokio unstables will need a `feature` as well, so I don't think it hurts to unconditionally set `--cfg tokio_unstable`

Jesse-Bakker (approved):

PR #2025, opened 2023-09-13
added temp_dir to LiveState

supergi0:
This will ensure the home directory does not get deleted until the `stop` function is called.

### DIFF
--- a/dozer-cli/src/live/state.rs
+++ b/dozer-cli/src/live/state.rs
@@ -38,9 +38,9 @@ struct DozerAndContract {
     contract: Option<Contract>,
 }
 
-pub struct ShutdownAndTempDir{
+pub struct ShutdownAndTempDir {
     shutdown: ShutdownSender,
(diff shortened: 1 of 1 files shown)


dozer-cli/src/live/state.rs
```diff
@@ -49,6 +50,7 @@ pub struct LiveState {
     run_thread: RwLock<Option<ShutdownSender>>,
     error_message: RwLock<Option<String>>,
     sender: RwLock<Option<tokio::sync::broadcast::Sender<ConnectResponse>>>,
+    temp_dir: RwLock<Option<TempDir>>,
```
chubei, line 53: Why we have to create another `RwLock`? Can we change `run_thread` to `RwLock<Option<ShutdownAndTempdir>>`?

```rust
struct ShutdownAndTempDir {
  shutdown: ShutdownSender,
  temp_dir: TempDir
}
```

chubei (approved):

PR #2026, opened 2023-09-13
fix: Fix checkpoint failure when using s3 storage

chubei:
Turns out s3 doesn't allow uploading a small part and then a big part. The "optimazation" doesn't work.

### DIFF
--- a/dozer-core/src/checkpoint/mod.rs
+++ b/dozer-core/src/checkpoint/mod.rs
@@ -1,11 +1,11 @@
-use std::{collections::HashMap, num::NonZeroUsize, sync::Arc};
+use std::{collections::HashMap, num::NonZeroUsize, ops::Deref, sync::Arc};
 
 use dozer_log::{
     camino::{Utf8Path, Utf8PathBuf},
     dyn_clone,
     replication::create_data_storage,
     storage::{self, Object, Queue, Storage},
-    tokio::{sync::mpsc::error::SendError, task::JoinHandle},
+    tokio::task::JoinHandle,
 };
 use dozer_types::{
     bincode,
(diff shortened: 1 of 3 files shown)


karolisg (approved):

dozer-core/src/checkpoint/mod.rs
```diff
@@ -177,68 +169,38 @@ impl CheckpointFactory {
...
-
-        self.queue.create_upload(key.clone())?;
-        self.queue.upload_chunk(
-            key.clone(),
-            (source_states.len() as u64).to_le_bytes().to_vec(),
-        )?;
-        self.queue.upload_chunk(key.clone(), source_states)?;
-        self.queue.upload_chunk(key.clone(), data)?;
-        self.queue.complete_upload(key)?;
+        let data = bincode::serialize(&RecordStoreSlice {
```
Jesse-Bakker, line 180: This will use `serde::Serialize::serialize_seq` for `data`, which will encode elements of `data` one by one instead of just `memcpy`'ing.

You can work around that by either: implementing `Serialize` and `Deserialize` on `RecordStoreSlice`, and use `serialize_bytes` to serialize `data`, or use [serde-bytes](https://docs.rs/serde_bytes/latest/serde_bytes/) with its wrapper types or annotations

chubei: No wonder I felt slowdown!

Jesse-Bakker: This optimization does need bincode 2, which is currently in the RC stage

Jesse-Bakker: Of course you could go the manual route instead of bincode 2 and just copy the len and the buffer into a vec yourself.

chubei: I'll go for the rc!

PR #2030, opened 2023-09-13
feat: Add instance type to deployment command

karolisg:


dozer-cli/src/simple/cloud/deployer.rs
```diff
@@ -20,6 +20,7 @@ pub async fn deploy_app(
     client: &mut DozerCloudClient<TokenLayer>,
     app_id: &Option<String>,
     num_api_instances: i32,
+    instance_type: Option<String>,
```
chubei, line 23: Let's group these two to `InfraRequest`, otherwise the number of parameters will grow quickly.

chubei (approved):

chubei (approved):

PR #2032, opened 2023-09-13
 fix: make snowflake connector async

abcpro1:
Make the the blocking snowflake connector `async` in the most straightforward way, which is to move blocking code to a separate thread with `spawn_blocking`.

Alternatives considered to the current snowflake client in use:
- [odbc-api](https://lib.rs/crates/odbc-api): async support is limited. connections are still blocking, as well as getting schemas.
- [snowflake-api](https://lib.rs/crates/snowflake-api): relies on undocumented snowflake apis, which may or may not break in the future.
- [snowflake-connector](https://lib.rs/crates/snowflake-connector): still "Under heavy development".

### DIFF
--- a/dozer-ingestion/src/connectors/snowflake/connector/snowflake.rs
+++ b/dozer-ingestion/src/connectors/snowflake/connector/snowflake.rs
@@ -2,7 +2,8 @@ use std::time::Duration;
 
 use crate::connectors::snowflake::connection::client::Client;
 use crate::connectors::{
-    ConnectorMeta, ConnectorStart, SourceSchemaResult, TableIdentifier, TableInfo, TableToIngest,
+    ConnectorMeta, ConnectorStart, SourceSchema, SourceSchemaResult, TableIdentifier, TableInfo,
+    TableToIngest,
 };
 use crate::errors::ConnectorError;
 use crate::ingestion::Ingestor;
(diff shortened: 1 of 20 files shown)


dozer-ingestion/src/connectors/snowflake/connector/snowflake.rs
```diff
@@ -162,13 +173,28 @@ async fn run(
 
             info!("[{}][{}] Reading from changes stream", name, table.name);
 
-            consumer
-                .consume_stream(&stream_client, &table.name, ingestor, idx, iteration)
-                .await?;
+            consumer.consume_stream(&stream_client, &table.name, &ingestor, idx, iteration)?;
 
-            interval.tick().await;
+            std::thread::sleep(interval);
```
abcpro1, line 178: This isn't an exact replacement. I need to find a blocking `interval.tick()`.

chubei (approved):
Thanks for fixing the trait!

dozer-ingestion/src/connectors/snowflake/connector/snowflake.rs
chubei: Do we still need to sleep? I guess this `tick` was added to avoid blocking. @karolisg ?

karolisg: I think we need to sleep to avoid too frequent requests

chubei: Got it

PR #2045, opened 2023-09-16
chore: refactor cloud methods

v3g42:


chubei (approved):
PR looks good, but I don't like that we don't have access to log stream on client side.

v3g42:
> PR looks good, but I don't like that we don't have access to log stream on client side.

I don't like it either. We will have to create a bit more infra for having. access to log stream between multiple servers. 

This is a compromise for the time being.

PR #2061, opened 2023-09-20
fix: snowflake regression

abcpro1:
Start the query exec iterator to get a valid schema. The iterator is considered empty if the schema is `None`.


chubei (approved):

abcpro1:
@chubei we need integration tests for snowflake.

chubei:
I remember we have integration tests. We do all the github action setup. Is it not working?

abcpro1:
There are no tests which name begins with `test_connector_` in the snowflake connector. I found four tests that begin with `test_disabled_` in [dozer-ingestion/src/connectors/snowflake/tests.rs](https://github.com/getdozer/dozer/blob/3ee2bab776ead61e87267b3a68bbfcd2f29e60d3/dozer-ingestion/src/connectors/snowflake/tests.rs).

chubei:
@karolisg do you know what's happening with the snowflake tests?

karolisg:
> @karolisg do you know what's happening with the snowflake tests?

I remember we had problems setting credentials in github CI, so we disabled it.

PR #2075, opened 2023-09-22
chore: include simple connection info in AppResponse

duonganhthu43:


dozer-types/src/models/connection.rs
```diff
@@ -226,3 +226,21 @@ pub enum ConnectionConfig {
     /// In yaml, present as tag" `!Dozer`
     Dozer(NestedDozerConfig),
 }
+
+impl ConnectionConfig {
+    pub fn get_type_name(&self) -> String {
```
chubei, line 231: Can you help change [this snippet](https://github.com/getdozer/dozer/blob/main/dozer-cli/src/simple/build/contract/mod.rs#L110) to a call of this function?

chubei (approved):

dozer-types/src/models/connection.rs
v3g42: I would also use lower case for matching so we don't have to worry about case in various places.

PR #2078, opened 2023-09-22
refactor: Use `default` and `Option` in a more principled way

chubei:
This PR does 3 things:

- Remove the use of `#[serde(default = ...)]`.

This annotation makes the `Default` implementation of a struct different from a deserialized struct from no configuration, and can lead to subtle bugs (happened several times).

- Remove unnecessary `Option` over `enum`s. This was the limitation of `prost` and resulted in code bloat.

- Add `#[serde(deny_unknown_fields)]`


v3g42:
This is awesome. Thanks @chubei

v3g42 (approved):

PR #2089, opened 2023-09-26
fix: Close connections in gRPC server on shutdown

chubei:


Jesse-Bakker:
What is the difference between using the custom AsyncWrite implementation and `serve_with_shutdown`?

chubei:
`AddrStream` is not affected by the signal passed to `serve_with_shutdown`, which results in nasty bugs in live. After dozer is stopped, any connection to the stopped server is still alive and the HTTP requests over it are served. The client sees old data from last run.

Our custom implementation `AddrShutdownStream` will close the connection (by calling `AddrStream::pool_shutdown) when shutdown signal is received. And `serve_with_shutdown` will actually wait for connection being disconnected on shutdown, thus the shutdown is complete and graceful.

Jesse-Bakker:
Okay, I dove way too deep into the source code of tonic and hyper. The way we used to do it with an `AbortHandle` is definitely wrong, as you noticed, because `hyper` will spawn a new task for every connection and those tasks will just keep running.

`serve_with_shutdown` and `serve_with_incoming_shutdown` should (according to their implementation) gracefully shutdown connections (they pass a channel sender to the task and await the receiver closing) and only complete its future after all connections are drained. If that's not working, it's a bug in `hyper`.

chubei:
It works correctly, but we can't wait for all connections to go down. We need to close the connection from server side.

Jesse-Bakker:
Alright, I get it now :+1:

dozer-api/src/grpc/mod.rs
```diff
@@ -11,23 +11,143 @@ mod shared_impl;
...
 use http::{Request, Response};
+use hyper::server::conn::AddrStream;
 use hyper::Body;
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 use tower::{Layer, Service};
 
+use crate::shutdown::ShutdownReceiver;
+
+#[derive(Debug)]
+enum ShutdownAddrStream<F> {
```
Jesse-Bakker, line 36: This can be simplified a lot by making this
```rust
struct ShutdownAddrStream<F> {
    inner: AddrStream,
    shutdown: Option<F>,
}
```

dozer-api/src/grpc/mod.rs
```diff
@@ -11,23 +11,143 @@ mod shared_impl;
...
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        let this = Pin::into_inner(self);
+        if let Err(e) = this.check_shutdown(cx) {
+            return Poll::Ready(Err(e));
+        }
+
+        match this {
+            Self::Alive { inner, .. } => Pin::new(inner).poll_write(cx, buf),
```
Jesse-Bakker, line 120: If `let Poll::Ready(Ok(..)) = self.inner.poll_shutdown()`, I think this will try to write to a closed stream and error

chubei: That's the intention. We don't have choice other than to error out in this case right?

Jesse-Bakker: Yeah, I think it's alright to error out, but we should be careful to catch the error upstream so we don't, for example, crash the `live` server

chubei: It's handled by `hyper`/`tonic` or whatever framework code. As service implementors we won't see this error.

dozer-api/src/grpc/mod.rs
Jesse-Bakker: If you want to be clearer about the state, it can be
```rust
enum State {
    Alive{ shutdown: F },
    Shutdown
}

struct ShutdownAddrStream<F> {
    inner: AddrStream,
    state: State,
}
```

but i think just using an `Option` with maybe a short comment is plenty clear in this case

chubei: While I was making this change, I realized that we didn't pool `AddrStream` shutdown until completion, making the disconnecting not so graceful. See the latest commit.

dozer-api/src/grpc/mod.rs
```diff
@@ -11,23 +11,127 @@ mod shared_impl;
...
+                    Ok(())
+                }
+                Poll::Ready(Err(e)) => Err(e),
+                Poll::Pending => Ok(()),
+            },
+            ShutdownState::Done => Ok(()),
+        }
+    }
+
+    fn pool_impl<T>(
```
Jesse-Bakker, line 71: `pool` -> `poll`

Jesse-Bakker (approved):

dozer-api/src/grpc/mod.rs
chubei: Oh I always make this wrong

PR #2091, opened 2023-09-27
refactor: Use `version` instead of `deployment` in all APIs

chubei:


duonganhthu43:
Should we also include the label Version context for Prometheus data ?

chubei:
Right, let me do that in another PR.

duonganhthu43 (approved):

PR #2098, opened 2023-09-27
feat: make polling interval configurable in snowflake config

abcpro1:
In case a user wants to change the default 5 seconds interval.

### DIFF
--- a/dozer-types/src/ingestion_types.rs
+++ b/dozer-types/src/ingestion_types.rs
@@ -1,10 +1,14 @@
 use prettytable::Table as PrettyTable;
 use schemars::JsonSchema;
-use std::fmt::Debug;
+use std::{fmt::Debug, time::Duration};
 
 use serde::{Deserialize, Serialize};
 
-use crate::{node::OpIdentifier, types::Operation};
+use crate::{
+    helper::{deserialize_duration_secs_f64, f64_schema, serialize_duration_secs_f64},
+    node::OpIdentifier,
+    types::Operation,
+};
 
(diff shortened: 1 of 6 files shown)


dozer-types/src/ingestion_types.rs
```diff
@@ -184,6 +190,15 @@ pub struct SnowflakeConfig {
...
     pub role: String,
+
+    #[serde(
+        default = "default_snowflake_poll_interval",
+        skip_serializing_if = "equal_default_snowflake_poll_interval",
+        deserialize_with = "deserialize_duration_secs_f64",
+        serialize_with = "serialize_duration_secs_f64"
+    )]
+    #[schemars(schema_with = "schema_for_duration_secs_f64")]
+    pub poll_interval_seconds: Duration,
```
chubei, line 201: Let's use `Option<Duration>` (or `Option<f64>`, with all the `...with` it seems easier) instead.

The reason behind this is that we've experienced several bugs because of inconsistent `Default` value with the default deserialized value. Therefore we decided to remove all `#[serde(default = ...)]` in https://github.com/getdozer/dozer/pull/2078.

chubei (changes requested):

dozer-types/src/ingestion_types.rs
abcpro1: I changed it to `Option<Duration>`. I didn't use `f64` directly because it doesn't implement `Eq` and `Hash`.

chubei (approved):
Cool. Last time I encounter this I just used integer, but this works better.

PR #2101, opened 2023-09-28
Inline FieldType array in RecordRefInner

Jesse-Bakker:
This saves a pointer per recordref, which can be a significant memory
saving in the presence of many small records. This also makes the Arc
containing RecordRefInner a thin pointer, saving some more memory if
there are many references to a record.


dozer-recordstore/src/in_memory/mod.rs
```diff
@@ -338,13 +338,41 @@ impl FieldRef<'_> {
...
+    }
+}
+
+impl RecordRefInner {
+    fn new(fields: Vec<Field>) -> Self {
+        let field_types = fields.iter().map(|field| field.ty());
+        let field_types_layout = Layout::array::<Option<FieldType>>(field_types.len()).unwrap();
+        let values_size = size(field_types);
+        let values_layout = Layout::from_size_align(values_size, MAX_ALIGN).unwrap();
+        let (layout, values_offset) = field_types_layout.extend(values_layout).unwrap();
```
chubei, line 374: Looks like a call to `layout`?

chubei (approved):
Cool. I didn't realize `Arc<SliceWithHeader>` is 16 bytes. Really fat!

Jesse-Bakker:
> Cool. I didn't realize `Arc<SliceWithHeader>` is 16 bytes. Really fat!

Yeah, any pointer to a DST will become fat by including the object's size. That goes for references, raw pointers, but also smart pointers like `Arc/Rc` and `Box`.
In this case, that size just moves into the allocation, because it has to, but that has the side benefit of shrinking the pointer size.

dozer-recordstore/src/in_memory/mod.rs
Jesse-Bakker: You're right :+1:

PR #2103, opened 2023-09-28
fix: Nested window function

mediuminvader:


chubei (approved):
@snork-alt just relalize it's you that requested review... I'm going to blindly merge.

PR #2111, opened 2023-09-29
feat: Read rocksdb path from config

karolisg:


Jesse-Bakker:
Should we by default create the temp dir in the `HomeDir`? Otherwise, if you don't specify a location in the config, you don't get any of the benefits of using rocksdb, which can be quite a footgun for users

PR #2117, opened 2023-10-02
fix: reduce memory consumption

abcpro1:


Jesse-Bakker (approved):

abcpro1:
The test case failure appears to be related to checkpointing. I will fix that, then we can merge.

PR #2135, opened 2023-10-09
feat: Use bytes instead of string as file content

karolisg:


dozer-cli/src/cloud/cloud_helper.rs
```diff
@@ -13,9 +14,17 @@ pub fn list_files(config_paths: Vec<String>) -> Result<Vec<File>, crate::errors:
 
         for entry in files_glob {
             let path = entry.map_err(CannotReadFile)?;
+            let f = fs::File::open(path.clone()).map_err(|e| CannotReadConfig(path.clone(), e))?;
+            let mut reader = BufReader::new(f);
+            let mut buffer = Vec::new();
+
+            reader
+                .read_to_end(&mut buffer)
+                .map_err(|e| CannotReadConfig(path.clone(), e))?;
```
chubei, line 23: [std::fs::read](https://doc.rust-lang.org/stable/std/fs/fn.read.html)?

chubei (approved):

PR #2138, opened 2023-10-09
chore: Update URL in hint text

chubei:


Jesse-Bakker:
We'll be changing the cloud URL to `cloud.getdozer.io`, which will obsolete this change

PR #2141, opened 2023-10-09
implement example for connectors

aaryaattrey:


dozer-types/src/ingestion_types.rs
```diff
@@ -331,13 +341,29 @@ impl S3Storage {
...
+    fn example() -> Self {
+        let s3_details = S3Details {
+            access_key_id: "<your_access_key_id>".to_owned(),
+            secret_access_key: "<your_secret_access_key>".to_owned(),
+            region: "<your_region>".to_owned(),
+            bucket_name: "<your_bucket_name>".to_owned(),
+        };
+        Self {
+            details: s3_details,
+            tables: vec![],
```
v3g42, line 354: Include one table as an example

dozer-types/src/ingestion_types.rs
```diff
@@ -331,13 +341,29 @@ impl S3Storage {
         )
     }
 }
+impl SchemaExample for S3Storage {
+    fn example() -> Self {
+        let s3_details = S3Details {
+            access_key_id: "<your_access_key_id>".to_owned(),
```
v3g42, line 347: leave as an empty string instead. Easier to double click and replace.

dozer-types/src/ingestion_types.rs
```diff
@@ -414,3 +446,105 @@ pub fn default_buffer_size() -> u32 {
 pub fn default_snowflake_poll_interval() -> Duration {
     Duration::from_secs(60)
 }
+
+impl SchemaExample for MongodbConfig {
+    fn example() -> Self {
+        Self {
+            connection_string: "mongodb://<username>:<password>@localhost:27017/<database_name>"
```
v3g42, line 453: use the local defaults that are typically used in a local mongo.

v3g42: Also include a db_name as a sample.

v3g42 (changes requested):
@aaryaattrey As these are examples,I would rather put example values which closely resemble what is to be expected. Look at postgres as an example.

dozer-types/src/ingestion_types.rs
```diff
@@ -414,3 +453,104 @@ pub fn default_buffer_size() -> u32 {
...
+            adapter: Some("arrow".to_owned()),
+        }
+    }
+}
+
+impl SchemaExample for KafkaConfig {
+    fn example() -> Self {
+        Self {
+            broker: "{{broker_url}}".to_owned(),
+            schema_registry_url: Some("{{schema_registry_url}}".to_owned()),
```
v3g42, line 489: These two are still remaining.

dozer-types/src/ingestion_types.rs
```diff
@@ -414,3 +453,104 @@ pub fn default_buffer_size() -> u32 {
...
+        }
+    }
+}
+
+impl SchemaExample for DeltaLakeConfig {
+    fn example() -> Self {
+        Self {
+            tables: vec![DeltaTable {
+                path: "{{delta_table_path}}".to_owned(),
```
v3g42, line 498: Here as well.

dozer-types/src/ingestion_types.rs
aaryaattrey: Should I put an empty string here, or some default value?

v3g42: just leave an empty value if you are unsure.

v3g42 (approved):

PR #2142, opened 2023-10-09
feat: Add a config to disable pipeline checkpointing by default

chubei:
This PR adds a new flag `enable_app_checkpoints`. When disabled, `EpochManager` stops creating `CheckpointWriter`s, effectively avoids persisting the processor record store, all processors and the record writer. A new field `sink_persist_queue` is added to `EpochCommonInfo` to continue persisting the log even if checkpoints are disabled.
This flag defaults to `false`, which means the checkpoints are disabled.

### DIFF
--- a/dozer-cli/src/live/state.rs
+++ b/dozer-cli/src/live/state.rs
@@ -420,11 +420,9 @@ fn get_dozer_run_instance(
         None => {}
     };
 
-    let app = &mut dozer.config.app;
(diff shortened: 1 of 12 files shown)


Jesse-Bakker:
Currently vacuum is called on the in_memory record store while checkpointing. We should make sure that is still called, even when checkpointing is disabled

chubei:
@Jesse-Bakker Moved the call to `vaccum` to another place. See if it makes sense?

PR #2156, opened 2023-10-16
refactor: Include `id` in `Record` response

chubei:
Previously we only include record `version` in `message Record`, and the record `id` was returned in yet another `message RecordWithId`. It was because of legacy reasons that don't exist anymore.

In this PR, we include `id` in `message Record` and remove `message RecordWithId`. The same change is applied on both common and typed APIs.

We also made a small change to `ingest.proto`: instead of taking `Record` we now take `repeated Value`, because ingestion doesn't care about record id or version.


Jesse-Bakker:
The change to `ingest.proto` will require changes to some samples and documentation

chubei:
> The change to `ingest.proto` will require changes to some samples and documentation

Sure. That's what the `doc-update-needed` label is for.

chubei:
@jianxiangxun @Jesse-Bakker are we merging this?

PR #2158, opened 2023-10-16
feat: Store `SourceStates` in log and cache

chubei:
This PR is a preparation for solving the log data loss problem.

- We store `SourceStates` in cache, so when cache reconnects to a log, it knows whether it needs to roll back.
- We keep track of the `SourceStates` that was persisted when a `Log` is restored from checkpoint, so a reconnecting client knows where to roll back to if necessary.

### DIFF
--- a/dozer-api/src/cache_builder/mod.rs
+++ b/dozer-api/src/cache_builder/mod.rs
@@ -35,7 +35,7 @@ pub async fn build_cache(
     labels: LabelsAndProgress,
 ) -> Result<(), CacheError> {
     // Create log reader.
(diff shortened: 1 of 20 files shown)


Jesse-Bakker (approved):
For my understanding: what's the reason for storing log offset and source states in separate lmdb databases?

chubei:
I planned to store the source states as key value pairs in a database, so log position has to go to another database, otherwise there needs be some magic key for it. Later I found it too troublesome and went with simple bincode serialization but forgot to merge them into the same database.

PR #2162, opened 2023-10-18
fix: Handle app restart in cache builder

chubei:


dozer-api/src/cache_builder/mod.rs
```diff
@@ -1,295 +1,299 @@
...
-                    labels.push(SNAPSHOTTING_LABEL, snapshotting_str(snapshotting));
-                    increment_counter!(CACHE_OPERATION_COUNTER_NAME, labels);
+    pub fn run(
+        mut self,
+        runtime: Arc<Runtime>,
+        mut cancel: impl Future<Output = ()> + Unpin + Send + 'static,
+        operations_sender: Option<(String, Sender<GrpcOperation>)>,
+    ) -> Result<(), CacheError> {
+        loop {
+            let Some(read_one_result) = runtime.block_on(self.read_one_with_cancel(cancel)) else {
```
Jesse-Bakker, line 129: `runtime.block_on()` on the hot path will be slower than the previous solution that ran the log reading in a worker (https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#non-worker-future), which was a bit cleaner as well.

I think I'd like to invert control here, and have an async task which orchestrates cache building workers, swapping these tasks out when the log gets swapped. Cancellation would also be handled in the orchestrator task, as would swapping the `serving` cache.

The worker would get its ops through a channel, which has the added benefit of making the worker purely sync and not require a `runtime`.

dozer-api/src/cache_builder/mod.rs
```diff
@@ -1,295 +1,299 @@
...
+    labels: Labels,
+    write_options: CacheWriteOptions,
+) -> Result<Box<dyn RwCache>, ApiInitError> {
+    let (alias, cache_labels) = endpoint_meta.cache_alias_and_labels(labels.clone());
+    match cache_manager
+        .open_rw_cache(alias, cache_labels, write_options)
+        .map_err(ApiInitError::OpenOrCreateCache)?
+    {
+        Some(cache) => {
+            debug_assert!(cache.get_schema().0 == endpoint_meta.schema.schema);
```
Jesse-Bakker, line 217: I think these should be regular `assert!`s or `Err` results, as if the schema is inconsistent, we'll be serving inconsistent data. Opening caches isn't on the hot path, so perf overhead should not have an impact on overall perf.

dozer-api/src/cache_builder/mod.rs
```diff
@@ -1,295 +1,299 @@
...
-}
+impl CacheBuilder {
+    pub async fn new(
+        cache_manager: Arc<dyn RwCacheManager>,
+        app_server_url: String,
+        endpoint: &ApiEndpoint,
+        labels: LabelsAndProgress,
+    ) -> Result<(Self, Vec<u8>), ApiInitError> {
+        // Connect to the endpoint's log.
+        let mut client = InternalPipelineServiceClient::connect(app_server_url.clone())
```
Jesse-Bakker, line 58: Should we try to re-use the (re-)connect logic? We can use the cancellation future for time-outs.

dozer-api/src/cache_builder/mod.rs
chubei: There were two reasons to implement it like this:

1. The `LogReader` actually has an internal buffer so performance is not a concern. `read_one` is essentially a [channel `recv`](https://github.com/chubei/dozer/blob/fix/consistency/dozer-log/src/reader.rs#L125). The `read_log_task` before this PR was implemented before the `LogReader` was modified to its current state.
2. If `CacheBuilderState::update` actually results in a new `building` cache, the `LogOperation`s from the last `LogReader` must not be passed to `CacheBuilderState::process`, so here once we get a `LogOperation` successfully, it's immediately passed to `CacheBuilderState::process`.

I'm not sure if I understand the "cache building workers" correctly. There's only one worker here that calls `CacheBuilderState::process`. It takes `&mut self` so we can't really use multiple workers.

dozer-api/src/cache_builder/mod.rs
chubei: I'm not sure how to reuse reconnect code. We don't recreate `InternalPipelineServiceClient` now (or should we?). For now the reconnect happens inside tonic. We just call `InternalPipelineServiceClient::get_id` and etc again on network failure.

dozer-api/src/cache_builder/mod.rs
Jesse-Bakker: I was thinking something along the lines of (very much pseudocode):
```
tokio::spawn(async move {
    loop {
        let (rx, tx) = mpsc::channel(1);
        let cache = open_or_create_cache();
        tokio::spawn_blocking(|| build_log(rx, cache))
        while let Ok(msg) = read_one(){
             tx.send(msg)   
        }
        reconnect()
    }
})
```
So, control of which cache to build and which to serve is pulled outside of building the cache and the actual cache building happens in a purely sync task that doesn't know about there being multiple caches.

chubei: I think I misunderstood you. You're talking about let the `cache` drop in `build_log` task, and create a new cache and task upon reconnectin?

chubei: Above code can be written equivalently as following?

```
tokio::spawn_blocking(||
    loop {
        let cache = open_or_create_cache();
        while let Ok(msg) = tokio::block_on(read_one()) {
             build_one(cache, msg);
        }
        reconnect()
    }
)
```

This is indeed clearer. Let me do a quick refactor.

Jesse-Bakker (approved):

PR #2171, opened 2023-10-20
chore: Don't write to disk if `build --locked`

chubei:


Jesse-Bakker:
When building locked, there are still some things that the build needs to write to disk, such as the proto files. Does that still happen with this change?

chubei:
Oh I get it now. Closing.

PR #2172, opened 2023-10-20
chore: Remove legacy `build_name` and `descriptor_bytes` from `EndpointMeta`

chubei:
- `build_name` is not necessary because we always build to `v0001`.
- `descriptor_bytes` is not necessary because it's generated from schema.


Jesse-Bakker (approved):
Nice!

PR #2180, opened 2023-10-24
fix: Remove redundant `idx` in `ProcessorRecordStore`

chubei:
This field is not necessary and was not correctly maintained in `deserialize_and_extend`

### DIFF
--- a/dozer-recordstore/src/in_memory/store.rs
+++ b/dozer-recordstore/src/in_memory/store.rs
@@ -109,7 +109,6 @@ pub struct ProcessorRecordStoreDeserializer {
 struct ProcessorRecordStoreDeserializerInner {
     records: BTreeMap<usize, RecordRef>,
     record_pointer_to_index: HashMap<usize, usize>,
-    idx: usize,
 }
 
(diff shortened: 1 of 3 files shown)


dozer-recordstore/src/in_memory/store.rs
```diff
@@ -176,8 +174,11 @@ impl StoreRecord for ProcessorRecordStoreDeserializer {
     fn store_record(&self, record: &RecordRef) -> Result<(), RecordStoreError> {
         let mut inner = self.inner.write();
 
-        inner.idx += 1;
-        let idx = inner.idx;
+        let idx = inner
+            .records
+            .last_key_value()
+            .map(|(index, _)| index + 1)
+            .unwrap_or(0);
```
abcpro1, line 181: This seemed like a slight pessimization since `last_key_value` is O(log n), but it looks like `store_record()` is not used much on `ProcessorRecordStoreDeserializer`. I only found two indirect invocations when building the join operator (for creating two null records).

abcpro1 (approved):

PR #2196, opened 2023-10-30
feat: enable SQL queries on cache endpoints

abcpro1:
- [x] Make host and port for the pgwire server configurable
- [x] Fix value conversions
   - [x] Decimals
   - [x] Date and time
- [x] Pushdown query optimization
   - [x] Limit pushdown (Note, `datafusion` does not support offset pushdown; it pushes down `OFFSET 10 LIMIT 10` as `LIMIT 20` 
   - [ ] ~~Projection pushdown~~ (currently not supported by the cache query planner)
   - [x] Predicate pushdown
- [ ] Implement `pg_catalog` tables. Some Postgres GUI clients don't work without them.
- [ ] Dozer Cloud integration
- [x] Enable web browsers to use SQL
    - [x] REST `/sql` endpoint
- [ ] Authentication


chubei (approved):
I didn't carefully read the `sql/datafusion` and `sql/pgwire` implementation. Other parts look good!

PR #2202, opened 2023-11-01
feat: Implement `dozer run lambda`

chubei:
This PR implements the JavaScript lambda runtime using deno, and the `dozer run lambda` command.

An example based on [the mysql sample](https://github.com/getdozer/dozer-samples/tree/main/connectors/mysql) looks like this:

- dozer-config.yaml

```yaml
app_name: dozer-chinook-mysql
version: 1

connections:
  - name: chinook_mysql
    config: !MySQL
      url: mysql://root:mysql@localhost:3306/Chinook

sources:
  - name: invoice
    table_name: Invoice
    columns:
      - InvoiceId
      - CustomerId
      - InvoiceDate
      - Total
    connection: chinook_mysql

endpoints:
  - name: invoice
    path: /invoice
    table_name: invoice

lambdas:
  - !JavaScript
    endpoint: invoice
    module: ./lambdas/log.js
```

- lambdas/log.js

```javascript
export default console.log;
```

To test, first run `dozer run app`, then in another process `dozer run lambda`. You should see the operations logged to the screen.

This is a PoC with several things missing:

- [ ] App restart: App server restart is not handled.
- [ ] Runtime restart: Runtime is stateless. Upon restart, it runs every lambda from scratch.
- [ ] JS async support: We need to wait the promise if the lambda function is async.
- [ ] Hardcode: The `LogReaderOptions` are hardcoded now.
- [ ] TS (and maybe more) support: We need to somehow reuse the `TypeScriptModuleLoader`.

Also we don't run lambda on `dozer run` because it's only a PoC for now.


Jesse-Bakker:
Is this meant to support JS UDFs? If so, should the configuration live under `udf`?

Ergonomics-wise, can we make the file a JS module which exports functions and have the UDF configuration specify which functions to import from it?

chubei:
It's not UDF. It's change data functions over the Dozer app output.

I pondered over the idea of exporting the function, but decided this will be easier because user doesn't have to maintain the function name in two different places.
Or maybe we can use the default export. No preference here.

chubei:
Actually using exported function might be easier, so we don't have jump between Rust and JS during the lambda registration phase.

dozer-lambda/src/js/worker/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+async fn worker_loop(
+    mut runtime: JsRuntime,
+    mut work_receiver: Receiver<Work>,
+    lambdas: Rc<RefCell<LambdaHashMap>>,
+) {
+    loop {
+        match poll_fn(|cx| {
+            poll_work_and_event_loop(&mut runtime, &mut work_receiver, &lambdas.borrow(), cx)
```
Jesse-Bakker, line 211: I think this might explode memory usage by calling the lambda function many times before it has completed. Should we instead run the lambda function to completion before calling it again?

chubei: You're right. I'll add this to the list.

chubei:
Looks like the two deno runtimes are interfering with each other, making one of the tests fail.

Jesse-Bakker:
> Looks like the two deno runtimes are interfering with each other, making one of the tests fail.

Weird, because they should be running in entirely different processes, which don't even run at the same time.

chubei:
> make the file a JS module which exports functions

Not sure if this is possible with Deno. I searched the generated doc of `deno_core` and `v8` but didn't see how we can get the exports of a module.

Jesse-Bakker:
You can get the module's namespace object using [JsRuntime::get_module_namespace](https://docs.rs/deno_core/latest/deno_core/struct.JsRuntime.html#method.get_module_namespace), which just contains all the exported objects.

chubei:
Oh it's called namespace instead of exports. Thanks!

Jesse-Bakker:
We could or course also use a wrapper javascript module, which imports lambdas from the supplied module and registers them

chubei:
> use a wrapper javascript module

That would be a detour, as the registering is implemented in Rust.

chubei:
@aaryaattrey pushed a new commit that updates both the config file and the way lambda is defined. Please check the updated PR description.

aaryaattrey:
> @aaryaattrey I pushed a new commit that updates both the config file and the way lambda is defined. Please check the updated PR description.

Sure.

chubei:
The tests pass with last commit. Do we merge this?

Jesse-Bakker:
What ended up being the problem causing the test to fail?

chubei:
No idea. I didn't look into it.

dozer-lambda/src/js/tests.rs
```diff
@@ -0,0 +1,206 @@
...
+    tokio_runtime.block_on(test_lambda_runtime_impl(tokio_runtime.clone()));
+}
+
+async fn test_lambda_runtime_impl(tokio_runtime: Arc<tokio::runtime::Runtime>) {
+    let (app_url, app_server) = mock::start_mock_internal_pipeline_server().await;
+    let lambda_modules = vec![JavaScriptLambda {
+        endpoint: mock::mock_endpoint(),
+        module: "src/js/test_lambda.js".to_string(),
+    }];
+    let lambda_runtime = Runtime::new(tokio_runtime, app_url, lambda_modules, Default::default());
```
Jesse-Bakker, line 25: Can we add a layer of abstraction to `Runtime`/`Trigger` so that we can test it without setting up a whole `InternalPipelineServer`? I'd like it if we could just do `Trigger::add_lambda(op_receiver: spsc::Receiver, schema: Schema)`.

dozer-lambda/src/js/worker/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+        runtime: Arc<Runtime>,
+        trigger: Arc<Mutex<Trigger>>,
+        lambda_modules: Vec<JavaScriptLambda>,
+    ) -> Result<Self, Error> {
+        let (init_sender, init_receiver) = oneshot::channel();
+        let (work_sender, work_receiver) = channel(1);
+        let runtime_clone = runtime.clone();
+        let handle = runtime.spawn_blocking(move || {
+            let lambdas = Rc::new(RefCell::new(HashMap::new()));
+            let mut main_worker = MainWorker::bootstrap_from_options(
```
Jesse-Bakker, line 76: I think we want to initialize a `JsRuntime` ourselves here. The `MainWorker` provides many deno extensions, a lot of which we don't want if we want to properly sandbox lambdas (such as the fs events extension). See, for example, supabases edge runtime: https://github.com/supabase/edge-runtime/blob/b2092c646ca83a1a0810bce5d1476656e50374a4/crates/base/src/deno_runtime.rs#L206

We can always add more extensions if we deem them safe and useful, but removing extensions will break users' code.

Jesse-Bakker:
Can we put this behind a feature flag? It adds a lot of extra dependencies and therefore build time and running cargo test OOM's my 32GB system on the linking step.

dozer-lambda/src/js/tests.rs
chubei: If we add that, there's basically no code left in `Trigger` to be tested.

dozer-lambda/src/js/worker/mod.rs
chubei: The main reason I'm using `MainWorker` is for [`MainWorker::evaluate_module`](https://docs.rs/deno_runtime/latest/src/deno_runtime/worker.rs.html#442-464). Don't feel like copying that snippet around.

How about we use `PermissionsContainer` to control permissions?

I also believe `alow_all` is suitable here, as user should be able to do whatever they want when self-hosting. We can blacklist `fs` in cloud.

chubei:
> behind a feature flag

There's no extra dependency introduced in this PR. `deno_runtime` was introduced in the js connector. Do we want to feature flag theses two together?

dozer-lambda/src/js/worker/mod.rs
Jesse-Bakker: If we want to be able to await the results of lambda's, for the sake of durability, we can't allow using the `Worker` API or subprocesses. Besides that, I think DX is better if the capabilities when running locally are the same as on cloud

Jesse-Bakker (approved):
Great work!

chubei:
I'll resolve the conversations and merge?

PR #2203, opened 2023-11-02
chore: Bump version to 0.2.0

chubei:


Jesse-Bakker:
Updating the versions of all crates every release seems a hassle, seeing as they're all internal, unpublished crates. Should we just set all versions to "0.0.0" except for `dozer-cli`, like suggested in https://matklad.github.io/2021/08/22/large-rust-workspaces.html ?

If we want to do that, now seems like a good time.

chubei:
It's just find and replace in VSCode, so not a hassle for me. `dozer-log-js` and `dozer-log-python` also have released artifacts and need the version number. So I'd like to keep the way it is now.

Jesse-Bakker (approved):

PR #2213, opened 2023-11-07
chore: Move `deno` dependency to `dozer-utils` to be shared across multiple crates

chubei:


Jesse-Bakker:
Can we move this to a new crate? Otherwise compilation of anything that depends on `dozer-utils` is stalled until deno has finished compiling if the feature is enabled.

chubei:
@Jesse-Bakker updated

Jesse-Bakker (approved):

PR #2214, opened 2023-11-07
Reduce json memory

Jesse-Bakker:
- Use ijson to make json more compact in memory
- Update to bincode@2.0.0-rc.3.

Tested on the [Amazon Berkeley Objects (listings)](https://amazon-berkeley-objects.s3.amazonaws.com/archives/abo-listings.tar) dataset, loaded using mongodb. Memory usage decreased by 75%

### DIFF
--- a/dozer-types/Cargo.toml
+++ b/dozer-types/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 chrono = { version = "0.4.23", features = ["serde"] }
 serde = { version = "1.0.189", features = ["derive", "rc"] }
 serde_json = { version = "1.0.93", features = ["std"] }
+ijson = "0.1.3"
 rust_decimal = { version = "1.32", features = ["serde-str", "db-postgres"] }
 bincode = "1.3.3"
 ahash = "0.8.3"
@@ -38,6 +39,7 @@ tokio-postgres = { version = "0.7.7", features = [
 serde_bytes = "0.11.12"
 arbitrary = { version = "1", features = ["derive"], optional = true }
 schemars = "0.8.15"
+bson = "2.7.0"
 
 [build-dependencies]
 tonic-build = "0.10.0"
--- a/dozer-types/Cargo.toml
+++ b/dozer-types/Cargo.toml
@@ -12,7 +12,7 @@ serde = { version = "1.0.189", features = ["derive", "rc"] }
 serde_json = { version = "1.0.93", features = ["std"] }
 ijson = "0.1.3"
 rust_decimal = { version = "1.32", features = ["serde-str", "db-postgres"] }
-bincode = "1.3.3"
+bincode = { workspace = true, features = ["serde"] }
 ahash = "0.8.3"
 thiserror = "1.0.48"
 parking_lot = "0.12"
@@ -39,7 +39,7 @@ tokio-postgres = { version = "0.7.7", features = [
 serde_bytes = "0.11.12"
 arbitrary = { version = "1", features = ["derive"], optional = true }
 schemars = "0.8.15"
-bson = "2.7.0"
+rmp-serde = "1.1.2"
 
 [build-dependencies]
 tonic-build = "0.10.0"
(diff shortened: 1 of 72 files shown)


dozer-types/src/types/field.rs
```diff
@@ -536,18 +737,15 @@ impl Field {
...
-            Field::Text(t) => Some(JsonValue::String(t.to_owned())),
-            Field::Null => Some(JsonValue::Null),
+            Field::UInt(u) => Some((*u).into()),
+            Field::U128(u) => Some((*u as f64).into()),
+            Field::Int(i) => Some((*i).into()),
+            Field::I128(i) => Some((*i as f64).into()),
+            Field::Float(OrderedFloat(f)) => Some((*f).into()),
+            Field::Boolean(b) => Some((*b).into()),
+            Field::String(s) => json_from_str(s.as_str()).ok(),
+            Field::Text(t) => Some(t.into()),
```
chubei, line 747: Should this go through `json_from_str`?

dozer-types/Cargo.toml
```diff
@@ -38,6 +39,7 @@ tokio-postgres = { version = "0.7.7", features = [
 serde_bytes = "0.11.12"
 arbitrary = { version = "1", features = ["derive"], optional = true }
 schemars = "0.8.15"
+rmp-serde = "1.1.2"
```
chubei, line 42: Why we're using message pack and bincode together?

dozer-types/src/json_types.rs
```diff
@@ -1,253 +1,88 @@
...
-where
-    T: Into<JsonValue>,
-{
-    fn from(opt: Option<T>) -> Self {
-        match opt {
-            None => JsonValue::Null,
-            Some(value) => Into::into(value),
+    match (l.destructure_ref(), r.destructure_ref()) {
+        (DestructuredRef::Object(l), DestructuredRef::Object(r)) => {
+            for (l_k, l_v) in l {
```
chubei, line 59: [IObject](https://docs.rs/ijson/latest/ijson/object/struct.IObject.html#) preserves the insertion order, so `Ord` will not be well defined?

chubei (commented):
This is great! I didn't check code under `dozer-sql` (wasn't familar enough with it). The only problem I saw is the `Ord` implementation.

dozer-types/Cargo.toml
Jesse-Bakker: Bincode is not self-describing, which in turn means it can't express JSON. Practically, this means that bincode doesn't support serde's `deserialize_any()`, which JSON implementations need, as the data type of the serialized data is given by the actual contents of the serialized format.

In this case, switching to msgpack inside the bincode implementation was the easiest, as msgpack matches JSON quite closely, and is almost as fast and compact as bincode.

dozer-types/src/json_types.rs
Jesse-Bakker: Good point. I'll iterate left sorted instead, so it's well-defined

dozer-types/src/types/field.rs
Jesse-Bakker: I'm not sure. The previous implementation had 
```rust
            Field::String(s) => match JsonValue::from_str(s.as_str()) {
                Ok(v) => Some(v),
                _ => None,
            },
            Field::Text(t) => Some(JsonValue::String(t.to_owned())),
```
which makes sense in its own way. WDYT?

chubei: I'm not sure either. Let's keep the behaviour then.

dozer-types/src/json_types.rs
Jesse-Bakker: I did something inspired by `BTreeMap`'s `Ord` implementation and added a test

chubei (approved):

dozer-types/src/json_types.rs
Jesse-Bakker: Agreed

PR #2230, opened 2023-11-21
fix: `serde-v8` conversion functions don't work well with `serde_json::Value`

chubei:


Jesse-Bakker:
Should we use `ijson::IValue` here instead of `serde_json::Value`, as that is our in-mem json representation?

chubei:
> Should we use `ijson::IValue` here instead of `serde_json::Value`, as that is our in-mem json representation?

Yes! Let me change.

dozer-lambda/src/js/worker/mod.rs
```diff
@@ -46,10 +45,10 @@ impl Worker {
     }
 }
 
-fn create_record_json_value(field_names: Vec<String>, values: Vec<Field>) -> Value {
-    let mut record = Value::Object(Default::default());
+fn create_record_json_value(field_names: Vec<String>, values: Vec<Field>) -> JsonValue {
+    let mut record = JsonObject::new();
```
Jesse-Bakker, line 49: Personal preference:
```rust
let values = values.into_iter().map(field_to_json_value);
IObject::from_iter(field_names.into_iter().zip(values)).into()

```

Jesse-Bakker (approved):

PR #2240, opened 2023-11-27
refactor: use ijson in queries

chubei:
This is a step further to removing all `serde_json::Value`.


Jesse-Bakker:
I'm not sure if we want to get rid of `serde_json::Value` everywhere. `serde_json` will always still be required for actually parsing the json, and its json representation is more ergonomic than `ijson`'s.

chubei:
Makes sense. Closing.

PR #2242, opened 2023-11-30
feat: json_object support

chloeminkyung:
SELECT JSON_OBJECT('id: user.id', 'town: user.town', 'town_2: Bristol') as info FROM users

### DIFF
--- a/dozer-sql/expression/src/error.rs
+++ b/dozer-sql/expression/src/error.rs
@@ -75,6 +75,10 @@ pub enum Error {
     FailedToCalculateVincentyDistance(
         #[from] dozer_types::geo::vincenty_distance::FailedToConvergeError,
     ),
+    #[error("Fail to deserialize: {0}")]
+    FailedToDeserialize(
+        #[from] dozer_types::errors::types::DeserializationError,
+    ),
 
     #[error("Invalid like escape: {0}")]
     InvalidLikeEscape(#[from] like::InvalidEscapeError),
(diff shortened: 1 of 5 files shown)


Jesse-Bakker:
The syntax is a bit unusual. Can we use the SQL/JSON standard syntax instead?
Ex.:
```sql
select JSON_OBJECT('id': user.id, 'town': user.town, 'town_2': 'LA')
```
and/or
```sql
select JSON_OBJECT('id' VALUE user.id, 'town' VALUE user.town, 'town_2' VALUE 'LA')
```

On advantage is that there is a clear distinction between value literals `'LA'` and column references `user.id`

chloeminkyung:
> The syntax is a bit unusual. Can we use the SQL/JSON standard syntax instead? Ex.:
> 
> ```sql
> select JSON_OBJECT('id': user.id, 'town': user.town, 'town_2': 'LA')
> ```
> 
> and/or
> 
> ```sql
> select JSON_OBJECT('id' VALUE user.id, 'town' VALUE user.town, 'town_2' VALUE 'LA')
> ```
> 
> On advantage is that there is a clear distinction between value literals `'LA'` and column references `user.id`.
> 
> More info: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_OBJECT.html#GUID-1EF347AE-7FDA-4B41-AFE0-DD5A49E8B370

yes, we are aware of that - this is hacky way to implement this feature as fast as possible, not merging for main, of course. We should take one of your approach when we are implementing properly later on.

PR #2243, opened 2023-11-30
fix: use default_max_num_records in list REST endpoint

abcpro1:
Currently, the REST list endpoint `/` returns a maximum of 50 records, regardless of the configured default limit.


chubei (approved):

chubei:
@abcpro1 do we release a patch version for this?

PR #2244, opened 2023-11-30
fix: Decimal should be serialized as a JSON Number not JSON String

abcpro1:


dozer-types/src/json_types.rs
```diff
@@ -114,7 +114,13 @@ pub fn field_to_json_value(field: Field) -> JsonValue {
         Field::String(s) => s.into(),
         Field::Text(n) => n.into(),
         Field::Binary(b) => b.into(),
-        Field::Decimal(n) => n.to_string().into(),
+        Field::Decimal(n) => serde_json_to_json_value(
```
Jesse-Bakker, line 117: Instead of turning it into a string, this will make it lose precision. I think it's better to just turn it into a string.

abcpro1: Not sure I understand your first sentence. Did you mean turning into a number?

dozer-types/src/json_types.rs
```diff
@@ -222,8 +228,48 @@ pub fn serde_json_to_json_value(value: Value) -> Result<JsonValue, Deserializati
...
+pub fn lossless_f64_parse_opt(s: &str) -> Option<f64> {
+    if can_parse_losslessly_to_f64(s) {
+        s.parse().ok()
+    } else {
+        None
+    }
+}
+
+/// efficiently check if a decimal number can be parsed to an f64 without loss of precision
+fn can_parse_losslessly_to_f64(s: &str) -> bool {
```
Jesse-Bakker, line 241: This function is not correct. Simple counter-example: `"0.1"`, which is rounded to `0.1000000000000000055511151231257827021181583404541015625` when parsed as a float.

Jesse-Bakker (changes requested):

dozer-types/src/json_types.rs
abcpro1: It works fine. I added a test case to guard against regressions as well.

abcpro1: I don't need to test that because I don't use it that way. I only use this function to make sure the input number as a string remains the same when it makes the roundtrip back to string with serde_json::to_string(). Binary floating point quirks are not a concern. I'm not sure if rust_decimal deals with them either.

Jesse-Bakker: The thing is, once you turn Decimal(0.1) into a float, it's not 0.1 anymore. If you later want to turn it back into a decimal, you don't know whether it should become Decimal(0.1) or Decimal(0.1000000000000000055511151231257827021181583404541015625), because the internal representation of both is the same, which means you lose precision.

We should either say that json numbers are floats and make users cast them to strings if they don't want to lose precisions when converting to json, or make them strings and make people cast to float if they are okay with losing precision. We can't have it both ways

abcpro1: If it makes it back to Decimal it should be as a string to Decimal, not as an f64 to Decimal. The only thing I have tried to guarantee in this implementation is this.

abcpro1: The name matches exactly because when it makes it back to string, it's the same.

Jesse-Bakker: It is true that this implementation will only return `true` for one normalized string representation per float value. (e.g. it will return `true` for `0.1` but `false` for `0.100000000000000004`, which would have the same in-memory representation as a float). The name `can_parse_losslessly_to_f64` doesn't accurately describe these semantics for me, as the conversion does rounding to the nearest representable float. `can_roundtrip_through_f64_losslessly` is closer. This also needs a docstring explaining it.

abcpro1:
@chubei Can you give an opinion?

chubei:
I was not following the conversation. Let me read it now.

dozer-types/src/json_types.rs
Jesse-Bakker: I'm saying that I understand what this is supposed to do now, and that that behavior is fine, but that the name of the function and the docstring don't adequately describe that, evidenced by the fact that this discussion is so long. So what I'm asking is that you change the name of the function and adequately explain this in the docstring.

abcpro1: I am generally not good at naming things, so I'll admit that, but blocking a pull-request for a name is not something we should do in fast development pace.

abcpro1: I will do the rename as you suggested.

Jesse-Bakker (approved):

dozer-types/src/json_types.rs
chubei: I get it now. In that case the `unwrap` can panic? https://github.com/getdozer/dozer/pull/2244/files#diff-28be2470f2d9e69d931883cac1ae784dbd1e5044b64a93db4182deb8fd657a0eR123

chubei: That's the part I don't understand. We know that `INumber` cannot represent all `Decimal` values, so this `unwrap` must happen in some cases?

chubei: Nice!

PR #2247, opened 2023-12-05
fix: Don't retry indexing if key size is too long

chubei:
The key size error used to lead to dead loop and indexing cannot make progress.


abcpro1 (approved):

Jesse-Bakker:
We have both `String` and `Text` types. Should we have a size restriction on `String` and restrict creating `SortedInverted` indexes to `String`?

chubei:
> We have both `String` and `Text` types. Should we have a size restriction on `String` and restrict creating `SortedInverted` indexes to `String`?

That would be the way to go.

PR #2249, opened 2023-12-06
feat: partial predicate pushdown for LIKE on cache SQL layer

abcpro1:
Speed improvement is very significant. 380ms with predicate pushdown compared to 3s 540ms without, in a 280MB dataset of 230K records. (Note the predicate pushdown case, while faster, incorrectly returns 25699 records only, while the non-pushdown case correctly returns 62813 records).
The lack of projection pushdown in the cache means that all 230K records, with all fields included, must be read from LMDB, converted to JSON, converted from JSON to Arrow format, then fed to Datafusion for filtering. Predicate pushdown for LIKE demonstrably reduces this overhead (although the cache doesn't do substring search, which makes this comparison inconclusive).

This pull-request is marked as a draft because the implementation is only partially correct. The `$contains` filtering option of the cache currently doesn't support substring search, making it insufficient for correct pushdown of `LIKE %word%` as `"$contains": "word"`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3684,6 +3684,7 @@ dependencies = [
  "tower",
  "tower-http",
  "tracing-actix-web",
+ "unicode-segmentation",
  "uuid",
 ]
 
--- a/dozer-api/Cargo.toml
+++ b/dozer-api/Cargo.toml
@@ -58,3 +58,4 @@ datafusion-expr = "32.0.0"
 serde_json = { version = "1.0.108", features = ["arbitrary_precision"] }
 pgwire = "0.16.1"
 tempdir = "0.3.7"
+unicode-segmentation = "1.10.1"
--- a/dozer-api/src/sql/datafusion/predicate_pushdown.rs
+++ b/dozer-api/src/sql/datafusion/predicate_pushdown.rs
@@ -2,11 +2,12 @@ use std::sync::Arc;
 
 use datafusion::arrow::datatypes::{Decimal128Type, Decimal256Type, DecimalType};
 use datafusion::scalar::ScalarValue;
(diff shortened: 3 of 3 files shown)


chubei (approved):
I think we can merge it because it's a marginal improvement.

abcpro1:
@chubei it returns incorrect results for a query string including the filter `LIKE '%appl%'`. Datafusion, slow as it is, will consider the word "apple" a valid result, while the cache currently will miss it.

chubei:
> @chubei it returns incorrect results for a query string including the filter `LIKE '%appl%'`. Datafusion, slow as it is, will consider the word "apple" a valid result, while the cache currently will miss it.

Right I missed this case.

abcpro1:
> Right I missed this case.

I updated the description a bit to add some more information.

Jesse-Bakker:
There are multiple ways to support substring search in the cache. We can subdivide the problem into multiple distinct cases:

- Single-word search
  - Full-word (`field.contains("word")`)
  Simple inverted index using words as tokens
  - Prefix (`field.contains("wor\w*")`)
  Range search (key_min=`wor`, key_max=`wos`) in inverted index
  - In (`field.contains("\w*wor\w*")`)
  Index scan on inverted index
- Phrase search
  - Full phrase
  Full-string inverted index (may want to only index a fixed-length prefix and compare against primary index to confirm match)
  - Prefix search
  Range search on full-string inverted index (see above)
  - In
  Table scan OR tokenize query and look up in tokenized index (can be sped up by embedding token positions in index and using those).

For substring pushdown, we can specialize at the single-word prefix and IN and phrase full phrase, prefix and IN levels.

PR #2250, opened 2023-12-11
feat: handle breaking and non-breaking schema changes for MySQL connector

abcpro1:
Handle schema changes in MySQL connector.
Breaking schema changes stop the ingestion with an error. Non-breaking schema changes do not affect ingestion.

Breaking changes are:
- Dropping or renaming a database, table, or a column in a table that is being ingested
- Change in the data-type of a column that is being ingested that results in a different dozer field data-type.

Non-breaking changes are:
- Column reordering
- Addition of columns
- Dropping or renaming databases, columns, and tables that are not being ingested
- Superficial change in column data-type such that the dozer field data-type is unchanged

Implementation notes:
`sqlparser` is used for parsing `ALTER TABLE` SQL queries from MySQL Binlog to determine the type of schema change that occurred, and choose an appropriate action, based on whether the schema change is breaking or non-breaking. However, `sqlparser` fails to parse some `ALTER TABLE` statements because MySQL dialect support is incomplete. For this reason, there is a fall-back mode that reloads the entire schema and manually discovers changes by comparing the old schema with the new one. As reloading the entire schema is less efficient, it is delayed until it is required, which is when the subsequent data operation arrives. This way, only one schema refresh is done for multiple `ALTER TABLE` operations if they arrive sequentially, instead of doing a full schema refresh for each `ALTER TABLE` operation individually.

### DIFF
--- a/dozer-ingestion/mysql/src/lib.rs
+++ b/dozer-ingestion/mysql/src/lib.rs
@@ -1,6 +1,7 @@
 use dozer_ingestion_connector::dozer_types::{
     errors::types::DeserializationError,
     thiserror::{self, Error},
+    types::FieldType,
 };
 use geozero::error::GeozeroError;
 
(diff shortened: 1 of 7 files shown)


dozer-ingestion/mysql/src/lib.rs
```diff
@@ -47,4 +47,7 @@ pub enum MySQLConnectorError {
 
     #[error("Failed to fetch query result. {0}")]
     QueryResultError(#[source] mysql_async::Error),
+
+    #[error("Schema had a breaking change: {0}")]
+    BreakingSchemaChange(String),
```
chubei, line 52: I'd prefer if we use something like

```rust
enum BreakingSchemaChange {
    DatabaseDropped(String),
    ColumnRenamed {
        table_name: String,
        old_column_name: String,
        new_column_name: String,
    },
    ...
}
```

instead of a `String`.

chubei (approved):
LGTM

chubei:
Github won't let me post my review comments so I'll comment here.

There're three places where `DatabaseDropped` and `TableDropped` are constructed with unnecessary `format!`.

abcpro1:
Fixed!

PR #2255, opened 2023-12-12
Don't panic if object_store connector can't read a table

Jesse-Bakker:
Users will from time to time place files with an inconsistent schema
in their S3 buckets. Instead of panicking when that happens, log an
error.


abcpro1 (commented):
It looks to me that the schema of the first file is selected, and any subsequent file has a different schema is skipped, and the ingestion continues with the remaining files. Did I understand correctly?

Jesse-Bakker:
@abcpro1 yes, that's basically it. I think we want to completely rework this connector sometime soon, as it has quite a few sharp edges, but I think this should do the job for now.

abcpro1 (approved):

abcpro1:
I see. Is it OK to skip files?

PR #2262, opened 2023-12-21
Various fixes for CLI output

Jesse-Bakker:
- Log loaded config files
- Remove printing logo
- Rename Home directory to Data directory in user-facing messages
- Initiating -> Initializing

### DIFF
--- a/dozer-cli/src/main.rs
+++ b/dozer-cli/src/main.rs
@@ -3,7 +3,7 @@ use dozer_api::shutdown;
 use dozer_cli::cli::cloud::CloudCommands;
 use dozer_cli::cli::types::{Cli, Commands, ConnectorCommand, RunCommands, SecurityCommands};
 use dozer_cli::cli::{generate_config_repl, init_config};
-use dozer_cli::cli::{init_dozer, list_sources, LOGO};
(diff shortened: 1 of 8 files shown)


dozer-cli/src/main.rs
```diff
@@ -312,27 +295,23 @@ fn parse_and_generate() -> Result<Cli, OrchestrationError> {
...
+fn init_configuration(cli: &Cli, runtime: Arc<Runtime>) -> Result<(Config, Vec<String>), CliError> {
+    dozer_tracing::init_telemetry_closure(None, &Default::default(), || -> Result<_, CliError> {
+        let res = runtime.block_on(init_config(
+            cli.config_paths.clone(),
+            cli.config_token.clone(),
+            cli.config_overrides.clone(),
+            cli.ignore_pipe,
+        ));
+
+        match res {
```
karolisg, line 307: Can this be changed to a more readable block?
```
if let Ok(config) = res { 
  runtime.spawn(check_update());
}
res
```

Jesse-Bakker: `check_update()` actually hasn't worked for a while, and in general is not very useful IMO, so I just removed it altogether.

karolisg (approved):

PR #2266, opened 2023-12-29
feat: Snowflake sink

abcpro1:
Introduce the Snowflake sink as a storage destination for Dozer pipeline transformations.
A snowflake table is updated at a configurable interval with the latest values from Dozer.


dozer-log/src/reader/checkpoint.rs
```diff
@@ -56,6 +63,7 @@ impl CheckpointedLogReader {
             .await
             .map_err(|e| Error::DescribeStorage(endpoint.clone(), e))?
             .into_inner();
+        dozer_types::log::warn!("storage response: {storage_response:?}");
```
chubei, line 66: Looks like debug log?

dozer-sinks/snowflake/src/lib.rs
```diff
@@ -0,0 +1,452 @@
+use std::{thread, time::Duration, usize};
+
+use dozer_api::shutdown::ShutdownReceiver;
+use dozer_ingestion_snowflake::{
```
abcpro1, line 4: Currently uses `dozer_ingestion_snowflake` crate directly.

chubei (approved):

dozer-log/src/reader/checkpoint.rs
abcpro1: Indeed.

dozer-sinks/snowflake/src/lib.rs
abcpro1: Switched to `odbc-api` crate.

PR #2305, opened 2024-01-05
Optimize postgres snapshotting

Jesse-Bakker:
The match on type was not properly optimized, so manually make a lookup
table for data conversion routines.
Also pre-allocate the full fields vector.
These changes together improve throughput by about 9%.

As we probably want to move to a different method for snapshotting altogether,
feel free to close this PR, but it might be nice inspiration for possible 
optimizations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3984,6 +3984,7 @@ dependencies = [
  "rustls 0.21.9",
  "rustls-native-certs",
  "serial_test 1.0.0",
+ "tokio",
  "tokio-postgres",
  "tokio-postgres-rustls",
  "uuid",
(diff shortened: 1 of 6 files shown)


chubei (approved):
Looking good. Have you identified the bottleneck?

Jesse-Bakker:
Current slowest parts are allocations (both the fields vec and string storage) and tokio `mpsc::Sender::send`. Both could be alleviated by using batching (arrow StringArray stores strings as a contiguous array with offsets pointing to individual strings, for example, only requiring one allocation per batch, if the string lengths known).

chubei:
Great. I'll verify and implement the batching change

PR #2308, opened 2024-01-09
perf: Add Operation::BatchInsert

chubei:
Here we use the `BatchInsert` operation to minimize channel sending overhead.

Processors currently just process as if it's sequential `Insert`s.

Change: 7.41s -> 2.32s
Target: 1.89s

### DIFF
--- a/dozer-cli/src/pipeline/dummy_sink.rs
+++ b/dozer-cli/src/pipeline/dummy_sink.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, time::Instant};
 
 use dozer_cache::dozer_log::storage::Queue;
 use dozer_core::{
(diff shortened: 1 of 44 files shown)


Jesse-Bakker (approved):

dozer-cli/src/pipeline/dummy_sink.rs
```diff
@@ -51,7 +55,28 @@ impl Sink for DummySink {
...
+        connection_name: String,
+    ) -> Result<(), BoxedError> {
+        self.snapshotting_started_instant
+            .insert(connection_name, Instant::now());
+        Ok(())
+    }
+
+    fn on_source_snapshotting_done(&mut self, connection_name: String) -> Result<(), BoxedError> {
+        if let Some(started_instant) = self.snapshotting_started_instant.remove(&connection_name) {
+            println!(
```
abcpro1, line 69: Should `println` be replaced with one of the log macros?

chubei: Sure

PR #2322, opened 2024-01-16
feat: Continue mysql ingestion

### LINKED ISSUE #2280: MySQL resuming
MySQL should be similar to Postgres.

karolisg:
Closes #2280

### DIFF
--- a/dozer-ingestion/mysql/src/binlog.rs
+++ b/dozer-ingestion/mysql/src/binlog.rs
@@ -517,7 +517,7 @@ impl BinlogIngestor<'_, '_, '_> {
                 .handle_message(IngestionMessage::OperationEvent {
                     table_index: table.def.table_index,
                     op: op?,
-                    state: Some(encode_state(&bl))
+                    state: Some(encode_state(&bl)),
                 })
                 .await
                 .is_err()
--- a/dozer-ingestion/mysql/src/binlog.rs
+++ b/dozer-ingestion/mysql/src/binlog.rs
@@ -226,7 +226,6 @@ impl BinlogIngestor<'_, '_, '_> {
                     let query = query_event.query_raw().trim_start();
 
                     if query == b"BEGIN" {
-                        // If receiving side is closed, we can stop ingesting.
                         transaction_position = prev_position;
                         seq_in_transaction = 0;
                     } else if query.starts_with_case_insensitive(b"ALTER")
--- a/dozer-ingestion/mysql/src/binlog.rs
+++ b/dozer-ingestion/mysql/src/binlog.rs
@@ -8,6 +8,8 @@ use super::{
     conversion::{IntoField, IntoFields, IntoJsonValue},
     schema::{ColumnDefinition, TableDefinition},
 };
+use crate::state::encode_state;
+use dozer_ingestion_connector::dozer_types::log::debug;
 use dozer_ingestion_connector::{
     dozer_types::{
         json_types::{JsonArray, JsonObject, JsonValue},
@@ -30,6 +32,7 @@ use mysql_common::{
     },
     Row,
 };
+use std::cmp::Ordering;
 use std::{
     collections::{HashMap, HashSet},
     ops::Deref,
@@ -39,6 +42,7 @@ use std::{
 pub struct BinlogPosition {
     pub filename: Vec<u8>,
     pub position: u64,
+    pub seq_no: u64,
 }
 
 pub async fn get_master_binlog_position(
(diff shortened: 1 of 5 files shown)


abcpro1 (commented):
This should make CDC resumable. Snapshotting reusability can be done in the future.
I have left some comments, some are functional and others are refactorings. Once the functional issues are resolved, we can merge this.

dozer-ingestion/mysql/src/binlog.rs
```diff
@@ -210,15 +225,9 @@ impl BinlogIngestor<'_, '_, '_> {
...
-                        && self
-                            .ingestor
-                            .handle_message(IngestionMessage::SnapshottingStarted)
-                            .await
-                            .is_err()
-                    {
-                        // If receiving side is closed, we can stop ingesting.
-                        return Ok(());
+                    if query == b"BEGIN" {
+                        transaction_position = prev_position;
```
abcpro1, line 229: Is there a reason why there is a `transaction_position`?

dozer-ingestion/mysql/src/binlog.rs
```diff
@@ -434,14 +433,38 @@ impl BinlogIngestor<'_, '_, '_> {
...
 
                         let table = table_cache.get_table_details(table_index).unwrap();
-                        self.handle_rows_event(&rows_event, &table, tme).await?;
+                        let mut pos = BinlogPosition {
+                            filename: self.next_position.filename.clone(),
+                            seq_no: seq_in_transaction,
+                            position: transaction_position,
+                        };
+
+                        pos.seq_no = self
```
abcpro1, line 442: `seq_in_transaction` is no longer updated.

dozer-ingestion/mysql/src/binlog.rs
karolisg: We need to start from transaction begin statement, because otherwise, we cannot get table map event. 
https://github.com/getdozer/dozer/blob/8dfcdffd9118e020232798da69092ce4f69d2c32/dozer-ingestion/mysql/src/binlog.rs#L528

abcpro1: I see. Then it's more tied to a `TABLE_MAP_EVENT` than transaction begin?

karolisg: Yes. Of course, we can start from that event, but I think all ingestion messages from the same MySQL transaction need to have the same `position` in `state`.

dozer-ingestion/mysql/src/binlog.rs
karolisg: Thanks, removed it.

dozer-ingestion/mysql/src/binlog.rs
```diff
@@ -434,14 +433,32 @@ impl BinlogIngestor<'_, '_, '_> {
...
                 }
 
                 event_type => {
                     trace!("other binlog event {event_type:?}");
                 }
             }
+
+            if binlog_event.header().log_pos() > 0 {
+                prev_position = binlog_event.header().log_pos().into();
+            }
```
abcpro1, line 461: can this become `prev_position = self.next_position.position`?

abcpro1: To make it clearer it can be move right before this line https://github.com/getdozer/dozer/blob/44d5130cd01b836de597b19fbd8d8c4c041eecab/dozer-ingestion/mysql/src/binlog.rs#L187

abcpro1 (approved):

dozer-ingestion/mysql/src/binlog.rs
karolisg: Completely removed this `prev_position = self.next_position.position` part. 

I have done it to store `log_pos` of transaction start (btw, `log_pos` is 0 for first event in log). So now I just calculate  it using 
```
transaction_pos.position = (binlog_event.header().log_pos()
      - binlog_event.header().event_size())
      as u64;
```

Example of first two messages
```
INFO Ingestion starting at BinlogPosition { filename: [98, 105, 110, 108, 111, 103, 46, 48, 48, 48, 50, 53, 56], position: 9816006, seq_no: 12 }
 INFO Header: BinlogEventHeader {
    timestamp: 1704611411,
    event_type: FORMAT_DESCRIPTION_EVENT,
    server_id: 1,
    event_size: 121,
    log_pos: 0,
    flags: EventFlags(0x0),
}
 INFO Header: BinlogEventHeader {
    timestamp: 1705407102,
    event_type: QUERY_EVENT,
    server_id: 1,
    event_size: 95,
    log_pos: 9816101,
    flags: EventFlags(LOG_EVENT_SUPPRESS_USE_F),
}
```

So now, when parsing query event, we calculate start log_pos by `log_pos` - `event_size`

abcpro1 (approved):

PR #2323, opened 2024-01-16
aerospike sink

Jesse-Bakker:
- Add Aerospike sink, based on aerospike-client-c
- Destructors for aerospike sink


dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,983 @@
...
+#[derive(Debug)]
+struct Client {
+    inner: NonNull<aerospike>,
+}
+
+// The aerospike client API is thread-safe.
+unsafe impl Send for Client {}
+unsafe impl Sync for Client {}
+
+macro_rules! t {
```
chubei, line 97: Can this be changed to a function like

```rust
unsafe fn check_return_code(result: as_status, err: MaybeUninit<as_error>) -> Result<(), AerospikeError> {
 ...
}
```

dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,983 @@
...
+                    inner: NonNull::new_unchecked(inner),
+                };
+                this.connect()?;
+                this
+            }
+        };
+        Ok(this)
+    }
+
+    unsafe fn connect(&self) -> Result<(), AerospikeError> {
```
chubei, line 140: This looks like a safe function

dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,983 @@
...
+        }
+    }
+
+    unsafe fn put(
+        &self,
+        key: *const as_key,
+        record: *mut as_record,
+        policy: as_policy_write,
+    ) -> Result<(), AerospikeError> {
+        unsafe {
```
chubei, line 156: We don't need to put `unsafe` in `unsafe fn`?

chubei (commented):
Will continue reviewing tomorrow

dozer-sink-aerospike/src/lib.rs
Jesse-Bakker: I made a function `as_try`, used like
```rust 
as_try(|err| aerospike_connect(self.inner.as_ptr(), err))?;
```

dozer-sink-aerospike/src/lib.rs
Jesse-Bakker: Yup. I do think in general it's a good idea to put the actually unsafe parts into separate unsafe blocks, each with their own `SAFETY:` comment explaining why they're sound, but with a single-line function, that's of course not very useful, and those comments don't currently exist :P

dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,983 @@
...
+    message: String,
+}
+
+impl From<as_error> for AerospikeError {
+    fn from(value: as_error) -> Self {
+        let code = value.code;
+        let message = unsafe {
+            let message = CStr::from_ptr(value.message.as_ptr());
+            // The message is ASCII (I think?), so this should not fail
+            message.to_str().unwrap()
```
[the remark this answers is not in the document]
Jesse-Bakker: It's a null-terminated string (which is why we need to round-trip it through CStr). `CStr::to_str()` might fail if it's not valid UTF-8, but in this case I think it's guaranteed to be valid UTF-8, so the unwrap won't panic.

dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,983 @@
...
+    fn new(hosts: &str) -> Result<Self, AerospikeError> {
+        let mut config = unsafe {
+            let mut config = MaybeUninit::uninit();
+            as_config_init(config.as_mut_ptr());
+            config.assume_init()
+        };
+        unsafe {
+            // The hosts string should never contain inner null bytes, and if
+            // it does, the user is doing something very weird.
+            let hosts = CString::new(hosts).unwrap();
```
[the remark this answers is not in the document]
Jesse-Bakker: You're right. I've moved conversion to CString out of `Client`, as `Client` is supposed to be a thin wrapper that only "knows" C and moved it to `SinkFactory::build` to return the correct error.

dozer-sink-aerospike/aerospike-client-sys/build.rs
```diff
@@ -0,0 +1,71 @@
...
+    let output_dir = build_dir.join("out");
+    let lib_dir = output_dir.join("lib");
+    let include_dir = output_dir.join("include");
+    let make_flags = vec!["TARGET_BASE=out"];
+
+    let current_dir = env::current_dir().unwrap();
+    let source_dir = current_dir.join("aerospike-client-c");
+    cp_r(&source_dir, &build_dir);
+
+    let mut make = Command::new("make");
```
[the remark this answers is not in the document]
Jesse-Bakker: I think this is the generally accepted way to call `make` from a build script. It's not too bad IMO

Jesse-Bakker:
Workflow fails because it needs to check out submodules recursively

dozer-sink-aerospike/src/lib.rs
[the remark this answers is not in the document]
chubei: I get the `CStr` part. Was just wondering if it's `'static`.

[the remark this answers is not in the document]
Jesse-Bakker: I think it is static on the C side, but I think it's fine to clone it on the rust side to be safe, as it's not on the hot path

dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+                Field::Timestamp(v) => self.set_str_key(
+                    key,
+                    // Use a delayed formatting to RFC3339 so we don't have to allocate an
+                    // intermediate rust String
+                    v.to_rfc3339(),
+                    allocated_strings,
+                ),
+                // Date's display implementation is RFC3339 compatible
+                Field::Date(v) => self.set_str_key(key, v.to_string(), allocated_strings),
+                // FIXME: Not sure how time unit should be used, but I'm pretty sure ignoring it
```
chubei, line 403: The unit is a hint, so when we convert to external systems that have limited duration precision, we know if it's possible or not.

For example, if the external system only supports duration in milliseconds, and our unit is nanoseconds, we know we have to either report an error or lose precision.

dozer-sink-aerospike/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+    }
+}
+
+struct Operations(Vec<MaybeUninit<as_operations>>);
+impl Operations {
+    fn new(size: usize) -> Self {
+        Self(Vec::with_capacity(size))
+    }
+
+    fn next(&mut self, n_bins: usize) -> *mut as_operations {
```
chubei, line 727: I assume we cannot move `as_operations`, so this function is `unsafe` because it should only be called at most `size` times.

chubei (approved):
Wonderful

dozer-sink-aerospike/src/lib.rs
Jesse-Bakker: Alright, so in this case it's fine to ignore it, as we're pushing a full-resolution string anyway

PR #2328, opened 2024-01-17
Add aerospike sink multithreading

Jesse-Bakker:
- Fix aerospike sink build if gcov is not found
- Add multi-threading to aerospike sink

Quite diminished performance with 2 and 3 threads, but barely any hit for 1 thread
and steadily increased performance for up to `available_parallelism()` threads.


chubei (approved):
spmc channel is handy!

PR #2329, opened 2024-01-17
refactor: Asyncify `Source`

chubei:
In this PR we change `Source::start` to `async` and feed it a tokio channel instead of a crossbeam channel. The immediate benefit is the simplified implementation of `ConnectorSource` and the merge of `SourceSenderNode` and `SourceListenerNode` into one `SourceNode`, and one less channel in the pipeline for each source. The ultimate goal of this refactor is to receive all sources' data in one thread and coordinate the commits there.


Jesse-Bakker (approved):
Makes a lot of sense. I'm not sure what the performance differences might be between crossbeam channels and tokio channels, so we might want to benchmark that.

PR #2335, opened 2024-01-19
refactor: Respect source transaction boundaries

chubei:
This PR changes the mechanism of pipeline `Commit` message generation.

# Connector behaviour change

<del>- Connector's "state" used to be defined per table as [`HashMap<String, TableState>`](https://github.com/getdozer/dozer/pull/2335/files#diff-508b945a1302f37f2ba77c285e5abb547c97bb6ff3532ae78439d97082798da3L97), now it's a single [`SourceState`](https://github.com/getdozer/dozer/pull/2335/files#diff-508b945a1302f37f2ba77c285e5abb547c97bb6ff3532ae78439d97082798da3R81).</del>
- Connector is expected to send [`IngestionMessage::TransactionInfo(TransactionInfo::Commit)`](https://github.com/getdozer/dozer/pull/2335/files#diff-6a16ca1c53879aa66bd8bbee894cbba9e03622ec87004b2bf9736491578d3579R33) messages on transaction boundary.

# Pipeline behaviour change

`Commit` used to be generated whenever a source has ingested a certain number of operations, or certain time has passed. Now it's generated when all sources happen to be at a transaction boundary.

Note that if there are multiple sources and data keep flow in, it's possible that pipeline `Commit` is never generated, because when source A is at transaction boundary, source B isn't necessarily that.

@karolisg I need your review on the [postgres connector](https://github.com/getdozer/dozer/pull/2335/files#diff-15ae44feac854295f7013262d589f23d0efe7d6f4a301b55be642049613074b6).
@abcpro1 I need your review on the [mysql connector](https://github.com/getdozer/dozer/pull/2335/files#diff-ffa4484b7b17313ce7c6ade56c7b4fa5c65fe85fe2305cc9aada4687b4b4f6fd).


karolisg:
For mysql, I think we need to send a commit on `XID_EVENT`.

For example
Screenshot 2024-01-19 at 17 31 31

chubei:
> For mysql, I think we need to send a commit on `XID_EVENT`.

Can you help fix that?

karolisg:
> > For mysql, I think we need to send a commit on `XID_EVENT`.
> 
> Can you help fix that?

Sure

Jesse-Bakker:
In stateless replication, every DAG with be a path, so the case of commits never happening won't occur, right?

chubei:
> In stateless replication, every DAG with be a path, so the case of commits never happening won't occur, right?

Not if we start doing denormalization.

Jesse-Bakker (approved):

PR #2339, opened 2024-01-22
refactor: Rewrite README

chubei:
Any comment or editing is welcomed.


chubei:
- Comparison with fivetran and airbyte
- Reproduction steps for the benchmarks

PR #2340, opened 2024-01-22
Log config path when running dozer commands

aaryaattrey:


chubei:
I didn't see the logging code?

aaryaattrey:
> I didn't see the logging code?

Its already present [here](https://github.com/getdozer/dozer/blob/1f95b1fef10717988a167b43b8ca4164dcf2c310/dozer-cli/src/main.rs#L69). We were logging the file name earlier, in this PR we append the cwd to get the whole path.

Related PR : https://github.com/getdozer/dozer/pull/2262

Issue #2342, opened 2024-01-23
Aerospike resuming


chubei:

chubei:
https://github.com/getdozer/dozer/discussions/2343

PR #2344, opened 2024-01-24
refactor: Make source state source wide instead of per table

chubei:
Part of https://github.com/getdozer/dozer/issues/2342

This is one of the purposes of https://github.com/getdozer/dozer/pull/2335


chubei:
@karolisg I need your help to fix the mysql's `encode_state` and. `decode_state` functions.

Jesse-Bakker:
Can we make source state a `Record` instead? That would make it more easily inspectable on the sinks and every sink already knows how to write records

chubei:
> Can we make source state a `Record` instead? That would make it more easily inspectable on the sinks and every sink already knows how to write records

Good idea! I'll do that in another PR.

Jesse-Bakker (approved):

PR #2347, opened 2024-01-25
Added type filter in the event

MG190202:


dozer-api/src/grpc/typed/service.rs
```diff
@@ -374,7 +375,25 @@ fn on_event(
...
+                .as_enum_number()
+                .ok_or_else(|| Status::new(Code::InvalidArgument, "type must be an EventType"))
+        })
+        .transpose()?;
+    let filter = EndpointFilter::new(
+        schema,
+        match event_type {
+            Some(o) => o,
+            None => EventType::All.into(),
+        },
```
chubei, line 394:
```suggestion
        event_type.unwrap_or(EventType::All)
```

chubei (approved):
Looking good now! Only one more comment

PR #2353, opened 2024-01-29
feat: clickhouse sink

karolisg:


dozer-sink-clickhouse/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+dozer-core = { path = "../dozer-core" }
+dozer-types = { path = "../dozer-types" }
+dozer-log = { path = "../dozer-log" }
+dozer-recordstore = { path = "../dozer-recordstore" }
+clickhouse = { git = "https://github.com/getdozer/clickhouse.rs.git" }
+serde = { version = "1.0.193", features = ["derive"] }
```
chubei, line 14: Can we use `dozer_types::serde`?

chubei (approved):

dozer-sink-clickhouse/src/ddl.rs
```diff
@@ -0,0 +1,102 @@
...
+        let typ = match field.typ {
+            FieldType::UInt => "UInt64",
+            FieldType::U128 => "UInt128",
+            FieldType::Int => "Int64",
+            FieldType::I128 => "Int128",
+            FieldType::Float => "Float64",
+            FieldType::Boolean => "Boolean",
+            FieldType::String => "String",
+            FieldType::Text => "String",
+            FieldType::Binary => "Array(UInt8)",
```
Jesse-Bakker, line 87: This is incorrect if the source binary column is nullable, as `Nullable(Array(T))` is not allowed. In that case we should probably map NULL to an empty array (and maybe raise a warning on schema creation).

dozer-sink-clickhouse/src/schema.rs
```diff
@@ -0,0 +1,200 @@
...
+            .query(&format!(
+                "DESCRIBE TABLE {database}.{table_name}",
+                table_name = table.name,
+                database = table.database
+            ))
+            .fetch_all::<ClickhouseSchemaColumn>()
+            .await?;
+
+        for field in schema.fields {
+            let clickhouse_column = columns.iter().find(|column| column.name == field.name);
```
Jesse-Bakker, line 118: Let's get rid of some nesting here and do something like:
```rust
    let Some(clickhouse_column) = columns.iter().find(|column| column.name == field.name) else {
        return Err(ClickhouseSinkError::ColumnNotFound(field.name));
    };

    match field.typ {
        // etc...
    }
```

dozer-sink-clickhouse/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+            Field::String(v) => FieldWrapper::String(v),
+            Field::Text(v) => FieldWrapper::Text(v),
+            Field::Binary(v) => FieldWrapper::Binary(v),
+            Field::Decimal(v) => FieldWrapper::Decimal(v),
+            Field::Timestamp(v) => FieldWrapper::Timestamp(v),
+            Field::Date(v) => FieldWrapper::Date(v),
+            Field::Json(v) => FieldWrapper::Json(v),
+            Field::Point(v) => FieldWrapper::Point(v),
+            Field::Duration(v) => FieldWrapper::Duration(v),
+            Field::Null => FieldWrapper::Null(None),
```
Jesse-Bakker, line 148: If it's not nullable, we can't map to null right? Or is this for the case where there's a default value?

karolisg: Updated binary handling part. During creation, we will show a warning if the source column is nullable. Also, during the insertion of data, we will map `null` as `vec![]`

PR #2356, opened 2024-01-31
Added push_events flag to affect the common services

MG190202:


dozer-api/src/grpc/client_server.rs
```diff
@@ -104,6 +104,7 @@ impl ApiServer {
             cache_endpoints.clone(),
             operations_receiver.as_ref().map(|r| r.resubscribe()),
             default_max_num_records,
+            self.flags.push_events.unwrap_or(true),
```
chubei, line 107: Use the `default_push_events` function instead of `true`

dozer-api/src/grpc/common/service.rs
```diff
@@ -115,6 +117,10 @@ impl CommonGrpcService for CommonService {
         let query_request = parts.2;
         let access = extensions.get::<Access>();
 
+        if self.push_events == false {
+            Status::invalid_argument("GRPC not implemented error");
```
chubei, line 121: Let's use `UNIMPLEMENTED` instead of `InvalidArgument`

chubei (changes requested):
Please don't include commits from the last PR

PR #2357, opened 2024-01-31
Added_new push_events flag affect common services

MG190202:


dozer-api/src/grpc/common/service.rs
```diff
@@ -114,6 +117,9 @@ impl CommonGrpcService for CommonService {
         let extensions = parts.1;
         let query_request = parts.2;
         let access = extensions.get::<Access>();
+        if !self.push_events {
+            unimplemented!("GRPC not implemented.");
```
chubei, line 121: Use grpc `UNIMPLEMENTED`

chubei (changes requested):

chubei (approved):

PR #2375, opened 2024-02-05
refactor: Handle schema and types in aerospike

karolisg:


dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -158,105 +350,218 @@ impl Connector for AerospikeConnector {
...
+    {
         let mut fields = vec![Field::Null; columns_map.len()];
-        if let Some(pk) = columns_map.get("PK") {
-            fields[*pk] = match event.key.last().unwrap() {
-                None => Field::Null,
-                Some(s) => Field::String(s.to_string()),
-            };
+        if let Some((pk, _)) = columns_map.get("PK") {
+            fields[*pk] = key
+                .clone()
```
Jesse-Bakker, line 418: Because you checked the key's length above, I think this whole block can just become
```rust
fields[*pk] = Field::String(key.remove(3));
```
which doesn't allocate at all.

I think the `PK` field will always be filled, right? If it's not, the message is not well-formed and we should error, I think. Anyway, downstream we can't handle messages without a PK

dozer-types/src/models/connection.rs
```diff
@@ -208,13 +208,41 @@ fn get_sslmode(mode: String) -> Result<SslMode, DeserializationError> {
...
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Eq, PartialEq, Clone, Hash, Default)]
 pub struct AerospikeConnection {
     pub hosts: String,
     pub namespace: String,
     pub sets: Vec<String>,
     #[serde(default)]
     pub batching: bool,
+    #[serde(default)]
+    pub replication: ReplicationSettings,
+    pub schemas: Option<ConfigSchemas>,
```
Jesse-Bakker, line 220: `schema` should also be `#[serde(default)]` I think. Otherwise you have to specify something, even if you only use aerospike as sink

dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -401,11 +409,9 @@ async fn map_event(
         return Err(AerospikeConnectorError::InvalidKeyValue(key.clone()));
     }
 
-    let set_name = match key.clone().get(1) {
-        None => Err(AerospikeConnectorError::NoSetNameFindInKey(key.clone())),
-        Some(Some(set_name)) => Ok(set_name.clone()),
-        Some(None) => Err(AerospikeConnectorError::SetNameIsNone(key.clone())),
-    }?;
+    let [_, Some(ref set_name), _, Some(ref pk_in_key)] = key.clone()[..] else {
```
Jesse-Bakker, line 412: This checks the length, so you can skip checking it above. The key is owned, so let's avoid cloning it both here and at the start of this function.

dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -414,18 +420,7 @@ async fn map_event(
     {
         let mut fields = vec![Field::Null; columns_map.len()];
         if let Some((pk, _)) = columns_map.get("PK") {
```
Jesse-Bakker, line 422: This get should be infallible (e.g. just an indexing operation `columns_map["PK"]`). The existence of a primary key is a precondition of this function, and if it doesn't exist, we're in a very weird state and it's alright to just panic here.

karolisg: I think this one would be correct. It would only fail if user want to use PK in ingestion and defines that column.
```
if let Some((pk, _)) = columns_map.get("PK") {
    if let Some(pk_in_key) = pk_in_key {
        fields[*pk] = Field::String(pk_in_key.clone());
    } else {
        return Err(AerospikeConnectorError::PkIsNone(key.clone()));
    }
}
```

Jesse-Bakker: Yes, you're right

Jesse-Bakker (approved):

PR #2376, opened 2024-02-06
feat: dozer ui and dozer ui update

duonganhthu43:
updated CLI with 
`dozer ui`
`dozer ui update`

### DIFF
--- a/dozer-cli/src/main.rs
+++ b/dozer-cli/src/main.rs
@@ -1,18 +1,22 @@
 use clap::Parser;
 use dozer_api::shutdown;
 use dozer_cli::cli::cloud::CloudCommands;
-use dozer_cli::cli::types::{Cli, Commands, ConnectorCommand, RunCommands, SecurityCommands};
+use dozer_cli::cli::types::{
+    Cli, Commands, ConnectorCommand, RunCommands, SecurityCommands, UICommands,
+};
 use dozer_cli::cli::{generate_config_repl, init_config};
 use dozer_cli::cli::{init_dozer, list_sources};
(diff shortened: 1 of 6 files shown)


dozer-cli/src/main.rs
```diff
@@ -61,6 +66,25 @@ fn run() -> Result<(), OrchestrationError> {
...
 
+    // running UI does not require config to be loaded
+    if let Commands::UI(run) = &cli.cmd {
+        if let Some(UICommands::Update) = run.command {
+            runtime.block_on(
+                ui::downloader::fetch_latest_dozer_app_ui_code()
+                    .map_err(AppUIError::DownloaderError),
+            )?;
+            info!("Run `dozer ui` to see the changes.");
+            return Ok(());
```
karolisg, line 77: You can just `return Ok(())` after whole block

dozer-cli/src/ui/downloader.rs
```diff
@@ -67,6 +67,14 @@ pub async fn fetch_latest_dozer_app_ui_code() -> Result<(), DownloaderError> {
     .await
 }
 
+pub fn validate_if_dozer_app_ui_code_exists() -> Result<bool, DownloaderError> {
+    let directory_path = get_directory_path();
+    let file_path = Path::new(&directory_path)
+        .join(LOCAL_APP_UI_DIR)
+        .join("contents");
+    Ok(file_path.exists())
```
karolisg, line 75: Why do you need result when there is no error returned in function?

karolisg (changes requested):

karolisg (approved):

Jesse-Bakker (approved):

PR #2385, opened 2024-02-08
Remove record-store

MG190202:
For adding record-store changes- remove it and its usage


MG190202:
It has 2 uses in dozer-cli -> in pipeline/dummysink and in pipeline/log_sink, should i remove its usage completely from these 2 files as i dont know about the usage of DummySink and LogSink? Please help

chubei:
I think all usage can be removed?

MG190202:
> I think all usage can be removed?

Yes, i have removed most of it, only some things are left like RecordStore enum

PR #2392, opened 2024-02-12
feat: AS XDR rewind

mediuminvader:


dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -389,6 +429,8 @@ impl Connector for AerospikeConnector {
                                         name: name.clone(),
                                         typ: if name == "inserted_at" {
                                             FieldType::Timestamp
+                                        } else if name == "PK" {
```
Jesse-Bakker, line 432: PK are not necessarily UInt

dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -276,13 +314,15 @@ async fn event_request_handler(
     let event = json.into_inner();
     let state = data.into_inner();
 
+    debug!("Event data: {:?}", event);
```
Jesse-Bakker, line 317: Logging all events under `debug` is a bit verbose. Let's either make this `trace` or not log it at all

dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -546,20 +616,40 @@ async fn map_events(
...
+                serde_json::Value::String(s) => {
+                    fields[*pk] = Field::String(s.clone());
+                }
+                serde_json::Value::Number(n) => {
+                    fields[*pk] = Field::UInt(
+                        n.as_u64()
+                            .ok_or(AerospikeConnectorError::ParsingUIntFailed)?,
+                    );
+                }
+                _ => todo!("Throw error when key is not a string or number"),
```
Jesse-Bakker, line 649: This should be implemented

dozer-sink-aerospike/src/lib.rs
```diff
@@ -232,6 +233,28 @@ impl Client {
...
+        as_try(|err| {
+            let res = aerospike_info_any(
+                self.inner.as_ptr(),
+                err,
+                null(),
+                request.as_ptr(),
+                response as *mut *mut i8,
+            );
+            let response_string = CStr::from_ptr(*response);
+            println!("response: [{:?}] {:?}", res, response_string);
```
Jesse-Bakker, line 254: Should be removed

dozer-ingestion/aerospike/src/connector.rs
karolisg: Yeah, it can be other types. To handle that user should use schema.json.

Jesse-Bakker: By default, aerospike doesn't actually store the keys, so hitting this path will be rare anyway

karolisg: Yeah. It is more for testing purposes

Jesse-Bakker (approved):

PR #2394, opened 2024-02-13
feat: Stop ingestion after number of events in dummy sink

karolisg:


Jesse-Bakker (approved):
Very useful for testing and simple benchmarks :+1:

PR #2395, opened 2024-02-13
feat: Print average latency in dummy sink

karolisg:


dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -432,6 +436,18 @@ async fn map_events(
             }
         }
 
+        if let Some((pk, _)) = columns_map.get("inserted_at") {
```
Jesse-Bakker, line 439: `pk` is not really the primary key, right? Let's call it `i`, `idx` or `index` instead

PR #2397, opened 2024-02-14
fix: convert commit_scn and commit_timestamp to Option types

prabhuvaibhav:
Handling the case where `commit_scn` and `commit_timestamp` are NULL.


dozer-ingestion/oracle/src/connector/replicate/log_miner/mod.rs
```diff
@@ -132,7 +132,9 @@ fn parse_and_map(
         return Ok(None);
     };
     Ok(Some(match parsed.kind {
-        ParsedLogManagerContentKind::Commit => MappedLogManagerContent::Commit(parsed.scn),
+        ParsedLogManagerContentKind::Commit => MappedLogManagerContent::Commit(
+            parsed.scn.ok_or(Error::NullValue(String::from("SCN")))?,
```
chubei, line 136: NullValue is not created for this. We should "expect" in the parser once we know the op code is DML

chubei (changes requested):

Jesse-Bakker:
In which cases are these NULL? Because we use COMMITTED_DATA_ONLY, shouldn't these always be available unless there is a misconfiguration?

Jesse-Bakker:
So, these can be null for operations like creating custom types. I think we should add some filters on the `logmnr_contents` query to only get operations for the tables we're interested in. We won't (soon) support custom types, so we might want to error if they're used in the tables we're interested in (then we'd need to support commit_scn to be null, so we can use the results to create an error message).

PR #2399, opened 2024-02-15
Aerospike sink: resumability

Jesse-Bakker:
This implements resuming for the aerospike sink in the following way:

 - Write the OperationId of every op to the affected record
 - On commit, write the OperationId to a separate set asynchronously

When restarting, the commit OperationId is returned as the last operation id.
This might be behind (because aerospike semantics and because we write it asynchronously),
but we catch that by only applying operations if the record they apply to has
an OperationId that is smaller than the applied operation. This means the
commit OperationId is only there to get a reasonably tight lower bound on the
OperationId, as it's not needed for correctness.


Jesse-Bakker:
Superseeded by #2437

PR #2403, opened 2024-02-16
feat: Support aerospike batching

karolisg:


Jesse-Bakker:
The sinks currently kind of assume `BatchInsert` is only used by snapshotting, because replication will have updates and deletes as well. I see we currently assume all operations are inserts?

karolisg:
> Yes, that's possible. It might be more efficient if you have mostly inserts (for example, for oracle, batch inserts use a fast path), but if there is a mix, using non-batch inserts will be more efficient (because a batch insert will always round-trip to the database, and to maintain ordering, the previous mixed batch will have to be flushed to the db as well, but a mix of inserts, updates and deletes will combine all these operations into a single batch).
> 
> So, long story short, batch insert is mostly for insert-only workloads. For mixed workloads, batch insert is less efficient on the sink side

Got it. Using simple insert now.

PR #2408, opened 2024-02-19
chore: adapt UI for multiple tables in 1 sink

duonganhthu43:


dozer-cli/src/ui/app/state.rs
```diff
@@ -161,6 +161,11 @@ impl AppUIState {
                 .collect::<std::collections::HashSet<String>>()
                 .into_iter()
                 .collect();
+            let sinks = config.sinks.to_owned();
+            let mut sink_names: Vec<String> = vec![];
+            for sink in sinks {
+                sink_names.push(sink.name);
+            }
```
chubei, line 168:
```suggestion
            let mut sink_names: Vec<String> = config.sinks.iter().map(|sink| sink.name.clone()).collect();
```

PR #2419, opened 2024-02-22
feat: Support integer on oracle ingestion/sink

karolisg:


dozer-ingestion/oracle/src/connector/mapping.rs
```diff
@@ -51,7 +53,12 @@ fn map_data_type(
         match data_type {
             "VARCHAR2" => Ok(FieldType::String),
             "NVARCHAR2" => unimplemented!("convert NVARCHAR2 to String"),
-            "NUMBER" => Ok(FieldType::Decimal),
+            "INTEGER" => Ok(FieldType::I128),
+            "NUMBER" => match (precision, scale) {
+                (Some(precision), Some(0)) if precision >= 22 => Ok(FieldType::UInt),
```
Jesse-Bakker, line 58: If precision >= 22, it won't fit in a UInt. Furthermore, `NUMBER` is signed.

dozer-ingestion/oracle/src/connector/replicate/log_miner/mapping.rs
```diff
@@ -42,6 +42,12 @@ fn map_value(
         ))),
         (ParsedValue::String(string), FieldType::Decimal, _) => Ok(Field::Decimal(string.parse()?)),
         (ParsedValue::Number(number), FieldType::Decimal, _) => Ok(Field::Decimal(number)),
+        (ParsedValue::Number(number), FieldType::Int, _) => {
+            Ok(Field::Int(number.to_i64().unwrap()))
+        }
+        (ParsedValue::Number(number), FieldType::UInt, _) => {
+            Ok(Field::UInt(number.to_u64().unwrap()))
```
Jesse-Bakker, line 49: This unwrap might panic if the number is negative

dozer-sink-oracle/src/lib.rs
```diff
@@ -246,10 +246,10 @@ impl OracleSinkFactory {
         for field in &schema.fields {
             let name = &field.name;
             let col_type = match field.typ {
-                dozer_types::types::FieldType::UInt => "NUMBER",
-                dozer_types::types::FieldType::U128 => unimplemented!(),
-                dozer_types::types::FieldType::Int => "NUMBER",
-                dozer_types::types::FieldType::I128 => unimplemented!(),
+                dozer_types::types::FieldType::UInt => "INTEGER",
+                dozer_types::types::FieldType::U128 => "INTEGER",
```
Jesse-Bakker, line 250: Does `INTEGER` have a large enough precision? I thought it had precision `22`, which is only enough for about 70 bits, I believe

dozer-sink-oracle/src/lib.rs
```diff
@@ -709,8 +714,14 @@ mod tests {
                 ON (D."id" = S."id" AND D."name" = S."name")
                 WHEN NOT MATCHED THEN INSERT (D."id", D."name", D."content", D."__txn_id", D."__txn_seq") VALUES (S."id", S."name", S."content", S."__txn_id", S."__txn_seq") WHERE S.DOZER_OPKIND = 0
                 WHEN MATCHED THEN UPDATE SET D."content" = S."content", D."__txn_id" = S."__txn_id", D."__txn_seq" = S."__txn_seq"
-                WHERE S.DOZER_OPKIND = 1 AND COALESCE(S."__txn_id" > D."__txn_id" OR (S."__txn_id" = D."__txn_id" AND S."__txn_seq" > D."__txn_seq"), TRUE) = TRUE
-                DELETE WHERE S.DOZER_OPKIND = 2 AND COALESCE(S."__txn_id" > D."__txn_id" OR (S."__txn_id" = D."__txn_id" AND S."__txn_seq" > D."__txn_seq"), TRUE) = TRUE
+                WHERE S.DOZER_OPKIND = 1 AND (S."__txn_id" IS NULL
```
Jesse-Bakker, line 717: If `S.__txn_id IS NULL`, but `D.__txn_id IS NOT NULL`, we're in a weird state. That means that this was first replicated with a source that does send an OpIdentifier for records, and then for a source that doesn't.

dozer-sink-oracle/src/lib.rs
karolisg: From my understanding, oracle can guarantee precision up to 38.

dozer-sink-oracle/src/lib.rs
karolisg: Yeah, but not sure how to handle this

dozer-ingestion/oracle/src/connector/mapping.rs
```diff
@@ -51,7 +53,12 @@ fn map_data_type(
         match data_type {
             "VARCHAR2" => Ok(FieldType::String),
             "NVARCHAR2" => unimplemented!("convert NVARCHAR2 to String"),
-            "NUMBER" => Ok(FieldType::Decimal),
+            "INTEGER" => Ok(FieldType::I128),
+            "NUMBER" => match (precision, scale) {
```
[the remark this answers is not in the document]
karolisg: Right. Updated patterns.

dozer-sink-oracle/src/lib.rs
karolisg: Makes sense. Removed `S."__txn_id" IS NULL`

dozer-ingestion/oracle/src/connector/mapping.rs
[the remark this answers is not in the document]
Jesse-Bakker: `len(str(2**64)) == 20`
That means that if we want to guarantee that the number fits in a 64 bit integer, we should have precision <= 19. 
This is the opposite of the condition in the sink, where we're detecting if an integer fits into a destination column.

Jesse-Bakker (approved):
One small number change, otherwise LGTM

PR #2420, opened 2024-02-23
feat: Add CHR, NVL, DECODE, SUBSTR and REPLACE SQL function

mediuminvader:


dozer-sql/expression/src/scalar/string.rs
```diff
@@ -272,6 +272,200 @@ pub(crate) fn evaluate_to_char(
...
+pub(crate) fn evaluate_chr(
+    schema: &Schema,
+    arg: &mut Expression,
+    record: &Record,
+) -> Result<Field, Error> {
+    let value = arg.evaluate(record, schema)?;
+    match value {
+        Field::UInt(u) => Ok(Field::String((((u % 256) as u8) as char).to_string())),
+        Field::U128(u) => Ok(Field::String((((u % 256) as u8) as char).to_string())),
+        Field::Int(i) => Ok(Field::String((((i % 256) as u8) as char).to_string())),
```
Jesse-Bakker, line 284: The behavior for negative arguments is not well-defined. In oracle, a negative argument results in a `numeric overflow` error. Let's error on negatives.

Furthermore, this assumes ascii. We don't support non-utf8 character sets within dozer. For arguments in the range [0, 255], the correct behavior is obvious. Outside that, in rust, valid values would be  [0, 0x10FFFF], excluding [0xD800, 0xDFFF].

Let's keep the `mod 256` behavior, but error on negative values and warn on values larger than 256?

dozer-sql/expression/src/scalar/field.rs
```diff
@@ -0,0 +1,109 @@
...
+pub(crate) fn evaluate_decode(
+    schema: &Schema,
+    arg: &mut Expression,
+    results: &mut [Expression],
+    default: Option<Expression>,
+    record: &Record,
+) -> Result<Field, Error> {
+    let arg_field = arg.evaluate(record, schema)?;
+    let mut i = 0;
+    while i < results.len() {
```
Jesse-Bakker, line 49: We can use `slice::chunks_exact()` here to simplify the loop:
```rust
for &[coded, decoded] in results.chunks_exact(2) {
    if arg_field == coded.evaluate(record, schema) {
        return decoded.evaluate(record, schema);
    }
}

```

dozer-sql/expression/src/scalar/string.rs
```diff
@@ -272,6 +272,200 @@ pub(crate) fn evaluate_to_char(
...
+    arg: &mut Expression,
+    position: &mut Expression,
+    length: &mut Option<Box<Expression>>,
+    record: &Record,
+) -> Result<Field, Error> {
+    let arg_field = arg.evaluate(record, schema)?;
+    let arg_value = arg_field.to_string();
+
+    let position_field = position.evaluate(record, schema)?;
+    let position_result = position_field.to_i128();
```
Jesse-Bakker, line 383: A string of i64::MAX length is about 8 exabytes, I believe, so let's just use u64 here, which tends to be faster for computations. Below, it's converted to `usize` anyway.

dozer-sql/expression/src/scalar/string.rs
```diff
@@ -272,6 +272,200 @@ pub(crate) fn evaluate_to_char(
...
+                    argument: length_field,
+                });
+            }
+            length_result.unwrap()
+        }
+        None => arg_value.len() as i128,
+    };
+
+    let mut iter = arg_value.char_indices();
+    let (start, _) = iter.nth(position_value.try_into().unwrap()).unwrap();
```
Jesse-Bakker, line 410: This will panic if the indices fall outside of the string's length. If the start is outside the string's length, we should return an empty string, and if the end falls outside of the string, we should just until the end of the string.

So, something like:
```rust
arg_value[start.unwrap_or(arg.len())..end.unwrap_or(arg.len())]
```

dozer-sql/expression/src/scalar/string.rs
```diff
@@ -283,11 +477,12 @@ mod tests {
     fn test_string() {
         proptest!(
             ProptestConfig::with_cases(1000),
-            move |(s_val in ".+", s_val1 in ".*", s_val2 in ".*", c_val: char)| {
+            move |(s_val in ".+", s_val1 in ".*", s_val2 in ".*", c_val: char, i_val: i64) | {
                 test_like(&s_val, c_val);
                 test_ucase(&s_val, c_val);
                 test_concat(&s_val1, &s_val2, c_val);
                 test_trim(&s_val, c_val);
+                test_chr(i_val);
```
Jesse-Bakker, line 485: Adding this here will significantly decrease the amount of times the above tests run, because of the increased search space, with the same amount of cases.

dozer-sql/expression/src/scalar/string.rs
```diff
@@ -589,4 +784,74 @@ mod tests {
...
     }
+
+    fn test_chr(val: i64) {
+        let row = Record::new(vec![]);
+
+        // Field::String
+        let mut value = Box::new(Literal(Field::Int(val)));
+        assert_eq!(
+            evaluate_chr(&Schema::default(), &mut value, &row).unwrap(),
+            Field::String((((val % 256) as u8) as char).to_string())
```
Jesse-Bakker, line 795: The test is the same as the implementation, which means it doesn't really test anything

dozer-sql/expression/src/scalar/string.rs
mediuminvader: Agree on negatives, the documentation was not clear on that.
What you mean by "larger than 256"?

Jesse-Bakker: The oracle documentation mentions that for single-byte encodings, the effective value will be `mod 256`. For multi-byte encodings, like unicode encodings, that's not the case. For example, for UTF-8, like rust's strings, valid values are [0, 0x10FFFF], excluding [0xD800, 0xDFFF].

Jesse-Bakker (approved):

PR #2425, opened 2024-02-27
Fix: Revert oracle resumability and fix snapshot data insert

karolisg:


dozer-sink-oracle/src/lib.rs
```diff
@@ -201,8 +200,10 @@ impl OracleSinkFactory {
                     FieldType::String | FieldType::Text,
                     OracleType::Varchar2(_) | OracleType::NVarchar2(_),
                 ) => {}
-                (FieldType::UInt | FieldType::Int, OracleType::Number(precision, 0))
+                (FieldType::U128 | FieldType::I128, OracleType::Number(precision, 0))
```
Jesse-Bakker, line 203: This is incorrect. We are checking here that the output column is suitable for storing this data type. For i128/u128, that means we need precision to be at least 39. For UInt/Int, that means precision needs to be at least 20.

Jesse-Bakker (changes requested):

dozer-sink-oracle/src/lib.rs
karolisg: Got it. Fixed condition.

dozer-sink-oracle/src/lib.rs
```diff
@@ -246,10 +245,10 @@ impl OracleSinkFactory {
         for field in &schema.fields {
             let name = &field.name;
             let col_type = match field.typ {
-                dozer_types::types::FieldType::UInt => "INTEGER",
-                dozer_types::types::FieldType::U128 => "INTEGER",
-                dozer_types::types::FieldType::Int => "INTEGER",
-                dozer_types::types::FieldType::I128 => "INTEGER",
+                dozer_types::types::FieldType::UInt => "NUMBER(20)",
+                dozer_types::types::FieldType::U128 => "NUMBER(39)",
```
Jesse-Bakker, line 249: This will fail, because the maximum precision is 38. That basically means that we don't support 128 bit integers, because they won't fully fit in an Oracle number. We could try to support them anyway and error if they exceed the 39 decimal places precision.

karolisg: Right. Marking it as uinimplemented now. Later we will need to implement logic for it.

Jesse-Bakker (approved):

PR #2426, opened 2024-02-27
Reduce disk space usage in CI

Jesse-Bakker:
Disable debuginfo in dev profile and remove MongoDB sidecar container

We can also disable debuginfo with an env var just for CI, but I think the
debuginfo _generally_ does not add too much value for us, and reducing build 
times and target dir size locally is a quite nice side benefit.
The `rust-cache` action already disables incremental compilation on CI, so
no need to also do that manually.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -66,9 +66,6 @@ jobs:
       - name: Rust cache
         uses: swatinem/rust-cache@v2
         
-      - name: MongoDB in GitHub Actions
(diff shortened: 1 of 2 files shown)


Jesse-Bakker:
In my testing, before, we used 23GB. After this, it should be more like 10 GB (out of 14 GB total available).

There are some duplicate deps between the dozer-ingestion-* tests and the general tests. We could split those into separate workflows to stay below the standard runners limit of 14GB

PR #2429, opened 2024-02-27
Add Assertion changes

MG190202:


dozer-sink-aerospike/src/lib.rs
```diff
@@ -35,10 +35,8 @@ use aerospike_client_sys::{
     as_val_val_reserve, as_vector, as_vector_increase_capacity, as_vector_init, AS_BATCH_WRITE,
     AS_BIN_NAME_MAX_LEN,
 };
-use dozer_core::{
-    node::{PortHandle, Sink, SinkFactory},
-    DEFAULT_PORT_HANDLE,
-};
+use dozer_core::node::{PortHandle, Sink, SinkFactory};
+pub const DEFAULT_PORT_HANDLE: u16 = 0xffff_u16;
```
Jesse-Bakker, line 39: Why is this redefined here?

MG190202: actually, it is being used at 3 other places in the same file, so i redefined it here, because otherwise it was showing a warning that DEFAULT_PORT_HANDLE is not being used if i directly import it

PR #2446, opened 2024-03-07
chore: Reduce logs level

karolisg:


Jesse-Bakker:
I think it would still be useful to get the count at a lower level than the actual messages. Like, debug for the count and trace for the messages. I definitely agree logging the count is too verbose for info.

karolisg:
@Jesse-Bakker Added targets to log statements. We can use it like this
`RUST_LOG=info,aerospike_http_server=debug,oracle_sink=debug dozer run`

Jesse-Bakker (approved):

PR #2453, opened 2024-03-14
feat: Use temp table in oracle sink and improve logging in aerospike

karolisg:


dozer-sink-oracle/src/lib.rs
```diff
@@ -621,6 +707,11 @@ impl OracleSink {
             batch.append_row(&[])?;
         }
         batch.execute()?;
+
+        self.conn.execute(&self.merge_statement, &[])?;
+
+        self.conn.commit()?;
```
Jesse-Bakker, line 713: Committing here is possibly incorrect. Let's make the `commit` and/or `flush_batch` methods decide when to commit. The merge statement should therefore also only be executed there to prevent it running multiple times with overlapping data in the temp table.

Jesse-Bakker (approved):

PR #2461, opened 2024-03-20
feat: Support aggregation functions in clickhouse ddl

karolisg:


Jesse-Bakker:
I think creating complex table definitions is outside the scope of dozer. If you need aggregate function states and materialized views etc, you should create the table in clickhouse directly and not through the dozer sink

PR #2474, opened 2024-04-03
Allow for non-opid-emitting sources to be restartable.

Jesse-Bakker:
This is useful for the aerospike source, because it uses XDR to track
source offsets.

Additionally, the oracle sink can now be append-only and it does not
need additional metadata columns in that case. It will also not create a
metadata table if the source does not emit OpIdentifiers.


dozer-ingestion/aerospike/src/connector.rs
```diff
@@ -245,28 +245,26 @@ async fn ingestor_loop(
...
-            sender: message.sender,
-        };
-
-        // Propagate panic in the pipeline event processor loop.
-        operation_id_sender.send(pending_operation_id).unwrap();
-
         // Ignore the error, because the server can be down.
         for message in message.messages {
-            let _ = ingestor.handle_message(message).await;
+            let _ = ingestor.handle_message(message).await.unwrap();
```
chubei, line 251: If we `unwrap`, we don't have to write `let _ = `. And the comment above should be removed.

Issue #1659, opened 2023-06-20
Support for `IN` clause in streaming SQL


snork-alt:
## Description
Dozer streaming SQL does not currently support the `IN` operator. We need to extend support for `IN` with the following formats:

### Lookup from a static list of values
```
SELECT column_name(s)
FROM table_name
WHERE column_name IN (value1, value2, ...);
```
This is pretty straightforward. Whenever an `Operation` is received, the value must be looked up from the list of static values provided in the SQL. 

The behavior is similar to an equality condition. The following table explains how the `WHERE` operator handles messages:

- `INSERT`: The insert message is propagated if the WHERE condition is matched
- `UPDATE`: The previous and new values must be evaluated against the condition:
    - If both the previous value and current value do not match the condition, no action is taken
    - If both the previous value and current value match, the UPDATE is propagated downstream
    - If the previous value does not match, but the new values match, an INSERT is propagated downstream
    - If the previous value matches, but the new value does not match, a DELETE is propagated downstream
- `DELETE`: The insert message is propagated if the WHERE condition is matched

### Lookup from an inner `SELECT`:
SELECT column_name(s)
FROM table_name
WHERE column_name IN (SELECT STATEMENT);
```
This case is more complex as the values of the inner SELECT can change. This can follow a similar behavior of a JOIN operator. Whenever an INSERT message is received for the inner SELECT, for example, a lookup of the parent table must be performed to emit the values that were not previously matched. This behavior is the same one implemented in the JOIN. For such a case, the DAG should be constructed using a JOIN operator.

hi-rustin:
I will give it a shot.

snork-alt:
Note: this has already been taken up by @hi-rustin

hi-rustin:
> ### Lookup from an inner `SELECT`:
> ```
> SELECT column_name(s)
> FROM table_name
> WHERE column_name IN (SELECT STATEMENT);
> ```
> 
> This case is more complex as the values of the inner SELECT can change. This can follow a similar behavior of a JOIN operator. Whenever an INSERT message is received for the inner SELECT, for example, a lookup of the parent table must be performed to emit the values that were not previously matched. This behavior is the same one implemented in the JOIN. For such a case, the DAG should be constructed using a JOIN operator.

@mediuminvader Could you please give some tips about how to implement it?

[...]

PR #2495, opened 2026-05-20
feat(sql): support IN clauses in streaming SQL

NightVibes33:
/claim #1659

## Summary
- add SQL NULL semantics for static `IN` / `NOT IN` lists and mark the expression output nullable
- cover static `IN` list streaming semantics for non-matching inserts, false-to-true updates, true-to-true updates, true-to-false updates, and deletes
- lower `expr IN (SELECT one_column ...)` filters into an inner join against a deduplicated derived subquery, reusing the existing join/set processors instead of adding a parallel membership engine
- preserve additional `AND` predicates and reject invalid multi-column `IN` subqueries
- add scripted streaming coverage for matching rows, late-arriving membership rows, unqualified outer identifiers, retained predicates, duplicate inner rows, and duplicate-membership deletion
- add sqllogictest fixtures for static `IN`, `IN (SELECT ...)`, retained predicates, duplicate inner rows, membership deletes, outer-row updates, and late membership inserts

## Issue coverage
This covers literal `IN` lists in selection predicates, including the update/delete behavior described in the issue, and streaming `IN (SELECT ...)` subqueries with membership semantics.

The subquery path rewrites into Dozer's existing streaming inner join plus a deduplicated derived subquery. That keeps runtime behavior aligned with the established join/product pipeline while preventing duplicate right-side rows from duplicating left-side output.

## Verification
- `rustfmt --edition 2021 --check dozer-sql/expression/src/execution.rs dozer-sql/expression/src/in_list.rs dozer-sql/src/builder/mod.rs dozer-sql/src/expression/tests/in_list.rs dozer-sql/src/product/set/set_factory.rs dozer-sql/src/tests/builder_test.rs`
- `git diff --check`
- `git show --check --stat HEAD`

### DIFF
--- a/dozer-sql/expression/src/execution.rs
+++ b/dozer-sql/expression/src/execution.rs
@@ -436,7 +436,7 @@ impl Expression {
                 negated: _,
             } => Ok(ExpressionType::new(
                 FieldType::Boolean,
-                false,
+                true,
                 SourceDefinition::Dynamic,
                 false,
             )),
--- a/dozer-sql/expression/src/in_list.rs
+++ b/dozer-sql/expression/src/in_list.rs
@@ -12,14 +12,28 @@ pub(crate) fn evaluate_in_list(
     record: &Record,
 ) -> Result<Field, Error> {
     let field = expr.evaluate(record, schema)?;
+    if field == Field::Null {
+        return Ok(Field::Null);
+    }
+
     let mut result = false;
+    let mut saw_null = false;
     for item in list {
         let item = item.evaluate(record, schema)?;
+        if item == Field::Null {
+            saw_null = true;
+            continue;
+        }
+
         if field == item {
             result = true;
             break;
         }
     }
+    if !result && saw_null {
+        return Ok(Field::Null);
+    }
+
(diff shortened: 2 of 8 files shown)


NightVibes33:
Coverage update for commit `8e3061b`: added scripted streaming cases for direct membership filtering, membership arriving after the outer row, unqualified outer identifiers, retained `WHERE` predicates, and duplicate inner rows behaving like set membership.

NightVibes33:
Coverage update for commit `865ae40`: added a static `IN` streaming regression for the update/delete behavior from the issue text. It covers ignored non-matching inserts, false-to-true inserts, true-to-true updates, true-to-false deletes, and deletes for matching old rows.

NightVibes33:
Coverage update for commit `d416606d`: added product-level SQL logic fixtures for static `IN`, `IN (SELECT ...)`, retained `AND` predicates, duplicate right-side membership rows, membership deletes, outer-row updates, and late membership inserts. Also added a scripted regression proving duplicate subquery membership remains active until the final matching right-side row is deleted.

PR #2497, opened 2026-05-21
Support streaming SQL IN subqueries

### LINKED ISSUE #1659: Support for `IN` clause in streaming SQL

ProtonsAndElectrons:
/claim #1659

Closes #1659.

## Summary
- Add a streaming `IN` / `NOT IN` subquery selection operator backed by the existing join-style indexed state stores.
- Route `WHERE left_key IN (SELECT right_key FROM ...)` through the SQL builder as a two-input processor.
- Preserve WHERE transition behavior for left inserts, updates, deletes, and right-side membership changes.
- Add focused coverage for membership appearing/disappearing, left update transitions, `NOT IN`, and SQL pipeline construction.

## Validation
- `cmd /c set PROTOC=C:\Users\Mio\oss-bounty-work\tools\protoc-25.3-win64\bin\protoc.exe&& cargo test -p dozer-sql in_subquery -- --nocapture`

Earlier broader local validation on this branch also passed:
- `cargo test -p dozer-sql -- --nocapture`

AI-assisted implementation disclosure: I used Codex to prepare and validate this change.


ProtonsAndElectrons:
Local focused validation is green on 8d9549: cargo test -p dozer-sql in_subquery -- --nocapture passed all 4 targeted tests (selection::in_subquery::* plus 	ests::builder_test::test_pipeline_builder_with_in_subquery).

